PR #9, opened 2011-07-22
Flatten the directory layout

maxandersen:
As discussed on mail, if we remove the superfluous directory these samples will be easier to use as quickstarts on OpenShift Express.


pmuir:
Merged. In future, you should rebase your changes onto upstream/master before submitting :-)

maxandersen:
sorry about that ;) thanks!

PR #10, opened 2011-08-18
Add automated datasource deployments for mysql and postgresql

stuartwdouglas:
I have added a feature to the jboss-as maven plugin that allows it to automatically deploy datasources. I think it would be a good idea to have a sample of this functionality in the quickstarts, to enable people to get started with a custom datasource quickly on AS7.

This pull request has an example pom that can automatically add mysql and postgresql datasources (and xa-datasources).

### DIFF
--- a/datasource/pom.xml
+++ b/datasource/pom.xml
@@ -9,6 +9,16 @@
     <packaging>pom</packaging>
     <name>JBoss AS Quickstarts: Datasources</name>
 
+    <description>
+        The pom provides example datasource configurations that can be deployed using the jboss-as maven
+        plugin. To deploy a datasource modify the database properties below (or specify them using -D) and
+        run maven with the appropriate profile for the database you wish to connect to. For example, to deploy
+        a postgresql XA datasource you would run:
+
+        mvn install -Ppostgresql-xa
+
+    </description>
+
     <url>http://jboss.org/jbossas</url>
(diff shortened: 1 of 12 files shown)


pmuir:
Rather than add this as a new quickstart, I think it would make more sense to add this to an existing quickstart, probably the login quickstart, as profiles, and optional features.

We would need more docs for this too.

WDYT?

stuartwdouglas:
I don't know about it being part of another example, as there is not really any way to just run it without also installing the project (I don't think there is anyway, unless there is some  maven feature I don't know about).

I envisaged people using this in the real world as a sub project that just installs the datasource, rather than having it integrated into their main build pom. With that said though if you think it is better in the login example I am happy to go with that. 

I agree it needs more docs, I will try and get onto that this week. It would also be good to get oracle and mssql datasources, but I don't have them availible to test.

pmuir:
How about adding it as a sub module of the login quickstart then?

stuartwdouglas:
Ok, I will do that.

pmuir:
Stuart, did you do this?

stuartwdouglas:
No, and it will need a bit of work due to AS7 management API changes. I will revisit after 7.1 is out

pmuir:
I'm going to close this pull for now :-)

PR #19, opened 2011-12-11
Helloworld JSP

nelvadas:
Hello
I created the HelloWorld JSP with JSP & JSTL 1.2
The project reuses the business and DAO layers of the MemberRegistration template.


sgilda:
Elvadas,
Thanks for creating this quickstart! I reviewed it and have the following suggestion:

This appears to be a JSP version of the 'kitchensink' quickstart, not the 'helloworld' quickstart. In the readme.md file, you even say "This project rebuilds the presentation tier of the kitchensink quickstart app using JSP and JSTL instead of JSF features." The name of your application is very confusing since it is not a helloworld example. Could you please rename the quickstart from 'jboss-as-helloworld-jsp to 'jboss-as-kitchensink-jsp' and change the references in the readme.md file also? Thanks!

nelvadas:
Thanks sgilda. I will correct it ASAP +1.

sgilda:
Elvadas, I found another issue. If the kitchen sink example and this one are deployed at the same time, there is a conflict with the "Member" table name which results in quite a few errors and exceptions in the server log. Could you rename hte member table to something else, maybe "MemberJSP' or something else that's unique? 
I believe you need to make the change in 2 places:
1) Member.java
   Add the table name to the @Table annotation, for example
   @Table(name = "MemberJSP", uniqueConstraints = @UniqueConstraint(columnNames = "email")) 
2) Modify the import.sql file
Thanks!

nelvadas:
done!

sgilda:
Elvadas,

I ran into a few issues when I tried to test this application.

1) When I test this quickstart and enter an invalid value, for example, a telephone number that is too short, I get an error on the page "Transaction rolled back", but no indication what was wrong. With the kitchensink JSF example, a validation error message "size must be between 10 and 12" appears next to the Phone #. I can see the error in the log, but would it be possible to display the reason it failed to the users on the page so they know what they need to change?

2) Also, it fails with a "Unique index or primary key violation" the first time I run it after restarting the server and deploying the application. 
For example, I enter the following values 
     Name: abc, Email: def@a.com, Phone: 1223343455. 
I get 'Transaction rolled back" and in the server log, I see:
    Caused by: org.h2.jdbc.JdbcSQLException: Unique index or primary key violation: 
        "PRIMARY_KEY_F ON PUBLIC.MEMBERJSP(ID)"; SQL statement:
        insert into MemberJSP (email, name, phone_number, id) values (?, ?, ?, ?) [23001-145]
If I try again with the exact same values a 2nd time, it works.

3) I also saw some strange behavior at one point when I typed into the input fields. For example, in the Email, I typed "s", it flashed, then the "s" became selected (highlighted) and the next letter I typed, for example "g", overwrote and replaced the "s" instead of adding it. So I ended up with "g" instead of "sg". This was happening with all the input fields. I undeployed the application, restarted the server, then redeployed and the problem disappeared. Do you have any idea what was causing that problem?

wesleyhales:
Just a few comments from my side:
- JSP and EL versions are 2.2 under Java EE 6 (not 2.0)
- You should be consistent with your page directives and add them to index.jsp as you have done with the others. (i.e. <%@ page language="java" contentType="text/html; charset=ISO-8859-1"
  pageEncoding="ISO-8859-1"%>)
-  And we should be using UTF-8 for character encoding

Looks good otherwise!

sgilda:
Elvadas, let me know when you make the above changes.

nelvadas:
Thanks sgilda
I noted your comments and I will send a new  version ASAP.
Good afternoon.

2011/12/14 sgilda <
reply@reply.github.com

> Elvadas, let me know when you make the above changes.
> 
> ---
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/jbossas/quickstart/pull/19#issuecomment-3142580

## 

*
*

_Cldt_
_Elvadas NONO_
_06 61 21 20 38_

nelvadas:
new versions + responses to yours questions. :)

2011/12/14 elvadas nono nelvadas@gmail.com

> Thanks sgilda
> I noted your comments and I will send a new  version ASAP.
> Good afternoon.
> 
> 2011/12/14 sgilda <
> reply@reply.github.com
> 
> > Elvadas, let me know when you make the above changes.
> > 
> > ---
> > 
> > Reply to this email directly or view it on GitHub:
> > https://github.com/jbossas/quickstart/pull/19#issuecomment-3142580
> 
> ## 
> 
> *
> *
> 
> _Cldt_
> _Elvadas NONO_
> _06 61 21 20 38_

## 

*
*

_Cldt_
_Elvadas NONO_
_06 61 21 20 38_

nelvadas:
@wesleyhales Thanks for the remarks

@ sgilda, Concerning the Error Message Transaction was rolled back,
Since I am using JSP, I have to perform some validation either with js ( no good idea in that quickstart) or at the servlet level,
but some validations like Constraint integrity when occurs in the EJB/JPA environment throw the Transaction rolled back,
I try to propagate the Exceptions which is catched when persisting the entity but no success result, JPA PersistenceException hierrachy is unchecked Exception and it seems like the solution is to get the root cause of the Transaction rolle back ex. Do you have another idea about this subject?

sgilda:
It seems to be working better now. 
If I type one letter in for the name, for exampe "A", I get an error saying "Name can not be null" even though it clearly is not null.
The exception handling change does not appear to be working. I still only get the message "Transaction rolled back". After looking at the code, I can see that the exception is just being rethrown and not returned to or processed by the form. It appears that none of other quickstarts do error handling either, so I think you can remove the additional error handling you added. Sorry I troubled you with it. When you remove it, I will process your pull. 
Thanks!

nelvadas:
Thanks sgilda, I m working on the exception mgmt, I catch the exection from Entity persist method and propagate it to the service lalyer, the servlet add the exception message  " transaction rolled back"  to the request attribute before forwardin  the request.

but from what i read on forums it comes from the EJB trans mgt and couple with JPA, I will change message to       "xxxx required".

nelvadas:
oupsssss! the validations constrainst is fake, I will check it,            request.getParameter (""").length <=1

sgilda:
Like I said, none of the other quickstarts do any additional error 
handling, so you don't need to with yours either. It's your call though. :-)

sgilda:
Wesley,  are satisfied with the changes? If so, I'll push them to jbossas.

sgilda:
Never mind. Still waiting for validation fixes. MemberRegistrationServlet is still testing for name length less than or equal to 1, meaning a one letter name like "A" gives an error that it's null:
if((value=request.getParameter("name")).length()<=1) {
    errorMessage.append("Name can not be null\n");
    }

nelvadas:
Sorry sgilda, My computer was out all the week end. I lost my battery charger. It s ok now :)
I try to fix catch and display exception properly. But i think it is not complete, I am working to see how to sole this issue using nested transaction  so that the transaction be rolled back in the DAO layer not in the business service caller (servlet ) first.
Good day:)

sgilda:
I don't have much GUI/JSF experience, but I played around with this a 
little in my own repository, https://github.com/sgilda/quickstart.

If you look at the register method in 
kitchensink-ml/src/main/java/org/jboss/as/quickstarts/kitchensink/controller/MemberRegistration.java, 
you'll see it catches specific exceptions, then adds them to the UI 
component that was added to the JSF page. I don't know if something like 
this would work with JSPs.

I don't know if code like this is even a best practice. Wesley, can you 
tell me if this the right way to process errors? ;-)

nelvadas:
ok thanks sgilda, I will check! but for me  with kitchensink-ml   the presentation tier and the DAO are not separated in ex management, DAO is aware of Facelet objects. very tigh coupling

sgilda:
That's what I was afraid of. Thanks.

pmuir:
sgilda, nelvadas, how are we doing on this? Close to getting it merged?

sgilda:
Still waiting for Elvadas to fix the validation code. See comment above 
"oupsssss! the validations constrainst is fake, I will check it, 
request.getParameter (""").length <=1"

nelvadas:
Already done sgilda , validations are ok and comited 

Envoyé depuis mon téléphone Nokia
-----Message d'origine-----
De : sgilda
Env.:  23/12/2011 14:47:10
Objet :  Re: [quickstart] Helloworld JSP (#19)

Still waiting for Elvadas to fix the validation code. See comment above 
"oupsssss! the validations constrainst is fake, I will check it, 
request.getParameter (""").length <=1"

On 12/23/2011 08:43 AM, Pete Muir wrote:

> sgilda, nelvadas, how are we doing on this? Close to getting it merged?
> 
> ---
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/jbossas/quickstart/pull/19#issuecomment-3260943

---

sgilda:
Sorry. Last time I checked it was still had validation logic for a name 
length <=1.

sgilda:
Looks good! The error messages are now being displayed so you know what's wrong. Eventually it would be nice to parse and display only the interpolated message from the entire exception, but this looks good for now. Thanks. I will try to merge it in the next few days.

pmuir:
Sande merged this.

PR #21, opened 2011-12-13
Change login example to be the same as Spring example

pmuir:
- rename to greeter
- remove login logic (anti-pattern)
- introduce DAO/Controller layer split to mirror Spring example


pmuir:
Sande, could you sanity check this change for me? Make sure I updated all relevant links..

sgilda:
Pete, 
I think you need to change the URL in the README from http://localhost:8080/jboss-as-login to http://localhost:8080/jboss-as-greeter.
When I grep 'login', I see the following instances. Not sure if you want to change them to 'greeter' also?
./src/main/resources/META-INF/persistence.xml:   <persistence-unit name="loginDatabase">
./src/main/webapp/greet.xhtml:      <h:form id="loginForm">
Should the following commented line be removed since the class doesn't exist?
./src/main/webapp/WEB-INF/beans.xml:

sgilda:
For some reason, the text that contained 'login' is missing from the comment above. I don't think it's necessary to rename all login to greeter, but just so you're aware:
./src/main/resources/META-INF/persistence.xml:   (greeterDatabase?)
       <persistence-unit name="loginDatabase">
./src/main/webapp/greet.xhtml:     (greeterForm?)
       <h:form id="loginForm">

pmuir:
References updated.

sgilda:
Looks good. Do I just click on "Merge pull request"? (This is my first one and I want to be sure)

pmuir:
Nooooo! This produces ugly merge commits. I will talk you through it on IRC.

sgilda:
Merged the changes.

PR #24, opened 2011-12-17
Kitchensink-ml quickstart

sgilda:
The kitchensink-ml quickstart needs to be reviewed by an SME before pushing to jbossas.
It is based on the kitchensink, but I've changed a few things:
1: Added multiple language support. It currently supports English and French. I need someone with i18n experience to check that this is implemented in the correct way and is using best practices.
2. Enhanced the error reporting. For example, if you enter a duplicate key, you get the specific localised error.
I want someone to verify that the error handling is being done properly and the correct technique is being used to notify the page.
NOTE: the kitchensink-ear-ml quickstart is NOT ready yet.


pmuir:
I'm not sure who the SME is for this @ JBoss…

Maybe someone at Amentra?

sgilda:
Maybe we should just toss this quickstart. It may not be useful.

pmuir:
I think let's defer it for this round.

sgilda:
Agreed. I closed the pull request.

PR #27, opened 2011-12-22
Quickstart 22: ready for review

npfr:
Quickstart 22: EARs are still supported in EE6 - some users will still want the EJB JARs, potentially multiple WARs all wrapped up in an EAR 

Ready for review.


sgilda:
Paul, 
I had trouble finding this project and eventually figured out it's the one in your "quickstarts/ear" folder. Is that correct? 
I think the name is too generic. Could you rename it in a similar manner as your 'ejb-in-ear' quickstart? 
Maybe 'ejbjar-and-war-in-ear' or 'ejb-and-war-in-ear' or something else more descriptive so that it doesn't conflict with other EAR projects?
Thanks!

sgilda:
Paul, in the meantime I decided to test your 'ear' project. I can successfully run 'mvn clean package'. 

However, when I run 'mvn jboss-as:deploy' against jboss-as-web-7.0.2.Final, I get a BUILD FAILURE:
[ERROR] No plugin found for prefix 'jboss-as' in the current project and in the plugin groups [org.apache.maven.plugins, org.codehaus.mojo] available from the repositories [local (/home/sgilda/.m2/repository), central (http://repo1.maven.org/maven2)] -> [Help 1]

It does deploy successfully though. I can see the deployment in the server console and am able to access the quickstart in a browser "http://localhost:8080/jboss-as-ear/".

Do you know why I'm getting the above error message?

npfr:
@sgilda Thanks for taking a look.

The quickstart was originally called "ejbjar-and-war-in-ear" but this resulted in rather strange artifact IDs for the child poms:

jboss-as-ejbjar-and-war-in-ear-ejb
jboss-as-ejbjar-and-war-in-ear-web
jboss-as-ejbjar-and-war-in-ear-ear

Do I need to follow the naming convention:

jbos-as-<quickstart.name>-<child_pom.name>

I agree the name 'ear' is to generic for the quickstart. Could I rename the quickstart to "ejbjar-and-war-in-ear", but keep the artefact ids the same? I think I would be breaking your naming convention if I did?

npfr:
@sgilda 

All you need to do is run 'mvn clean package'. I was having trouble invoking the deploy plugin from the root pom, for just the ear pom. The problem I had was that maven failed because (I think) it couldn't find this plugin in the ejb and web poms.

I modified the lifecycle of the deploy plugin to run at the 'package' phase. On second thoughts, this is probably a hack. It doesn't make sense to do a deployment during the package phase.

I'm in the process of figuring out how to do the deploy only when the jboss-as:deploy goal is specified.

I've managed to avoid using maven until I (recently) joined JBoss, so you will need to excuse my lack of knowledge in this area ;-)

sgilda:
I agree that the suggested names are long and awkward, but I would 
prefer to keep some sort of naming convention. It will be less confusion 
for customers and there will be less chance that 2 quickstarts will 
deploy with the same web context.

Can you change it to ejb-in-ear. That is much shorter and consistent 
with your other quickstart, ejb-in-war. I think that is descriptive for 
the quickstart, don't you?

sgilda:
Yes, this does sound like a hack ;-).

I don't have much Maven experience, but I'll take a look at it. If we 
can't figure it out, we can ask for help from the Maven experts after 
the holidays.

npfr:
@sgilda 

I've renamed the quickstart to ejb-in-ear and also updated the artift ids in the poms.

With some help from @tomjenkinson I've got the deployment working outside of the package phase. You can now type:

mvn package jboss-as:deploy

to build and deploy from the root pom.

I think it's ready for another review.

Thanks,

Paul.

npfr:
@Sgilda, @pmuir 

I've applied the relevant comments from https://github.com/jbossas/quickstart/pull/26 to this quickstart.

ejb-in-ear/ear/pom.xml
```diff
@@ -0,0 +1,88 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+    <!--
+        This pom builds the EAR artifact and includes the ejb and web modules.
+    -->
+
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>org.jboss.as.quickstarts</groupId>
+    <artifactId>jboss-as-ejb-in-ear-ear</artifactId>
```
pmuir, line 12: I think we need to come up with a better name for this quickstart. What about giving it a non-descriptive name, or a name related to the domain, rather than the technology? Not sure...

pmuir:
Other than the insane-name issue, this looks good to me, so if Sande is happy one of us can merge it :-)

npfr:
Yeah, it's certainly a tricky one to name. Have you read through our naming discussion above?

sgilda:
Pete, can think of a good name for it that will make it unique and descriptive?

pmuir:
Not really, let's go with this for now.

pmuir:
Merged

PR #28, opened 2011-12-22
Quickstart #06: JSF, Richfaces

bleathem:


sgilda:
Brian,
This is great example! I do have a few comments and suggestions. 
-  In the README file, the header says 'Numberguess Example', but when you run the quickstart, It should be 'Helloworld JSF Example.
-  It would be nice if the README had a little more information about what is different about this example. Maybe more about what's going on in the code and the use of Rich Faces.
-  Could you add more information about what you should expect to see when you run it? I was a little thrown when I accessed the page. There was no page title or header and It was already filled in with a default name. It would be nice if the page had a header or title like "Helloworld using JSF and RichFaces" or something that describes the example. It would also be nice if it had instructions, like "Type a name in the input field and watch the following text change as you type. This is all done without a round-trip to the server..." or something like that to give them an idea of the benefit of the example.
  What do you think?

bleathem:
These are great suggestions Sande, I've updated the pull request taking your suggestion into account.

helloworld-jsf/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+The example can be deployed using Maven from the command line or from Eclipse using
+JBoss Tools.
+
+To set up Maven or JBoss Tools in Eclipse, refer to the _Getting Started Guide_.
+
+To deploy to JBoss AS 7, start JBoss AS 7, and type `mvn package jboss-as:deploy`. 
+The application is deployed to <http://localhost:8080/jboss-as-helloworld-jsf>. You
+can read more details in the _Getting Started Guide_.
\ No newline at end of file
```
pmuir: Just for consistency, could we align this readme to the others used. For example https://github.com/jbossas/quickstart/blob/master/kitchensink/readme.md

helloworld-jsf/pom.xml
```diff
@@ -0,0 +1,155 @@
...
+         is deployed -->
+      <finalName>jboss-as-helloworld-jsf</finalName>
+      <plugins>
+         <plugin>
+            <artifactId>maven-war-plugin</artifactId>
+            <version>2.1.1</version>
+            <configuration>
+               <!-- Java EE 6 doesn't require web.xml, Maven needs to catch 
+                  up! -->
+               <failOnMissingWebXml>false</failOnMissingWebXml>
```
pmuir: This comment and config is not needed as you are using a web.xml

helloworld-jsf/README.md
bleathem: Sure, I'll change it, but I _was_ being consistent with the REAME of the numberguess project:
https://raw.github.com/jbossas/quickstart/master/numberguess/README.md
The quickstart instructions suggest the numberguess example as a good example to follow:
https://docs.jboss.org/author/display/AS71/Writing+a+quickstart

Perhaps we should change the instructions, or update the README of numberguess.

pmuir: Yep, this is our mistake, we'll update it :-)

pmuir:
Generally looking great!

bleathem:
Suggested changes applied.

pmuir:
Thanks Brian. I squashed all your commits down to one, and merged it :-D

PR #29, opened 2011-12-23
Add reminder to update pom.xml

npfr:
When adding a new quickstart, are we supposed to also add it to the modules section of the root pom.xml?

If so, I think the README.md needs updating, to something similar to this pull request.


pmuir:
Thanks, squashed and merged.

PR #31, opened 2012-01-02
quickstart - singleton session bean

spagop:
here is my second usecase singleton session bean quickstart


helloworld-ssb/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+System requirements
+-------------------
+
+All you need to build this project is Java 6.0 (Java SDK 1.6) or better, Maven
+3.0 or better.
+
+The application this project produces is designed to be run on a JBoss AS 7.1.0 or EAP 6. 
+ 
+NOTE: The artifacts will come from the JBoss Community Maven repository, a superset of the Maven central repository.
```
pmuir: This is no longer true, and we do generally need to remove this line from the readmes. But we should make sure new ones don't have it!

helloworld-ssb/src/main/java/org/jboss/as/quickstarts/managedbeans/MyManagedBean.java
```diff
@@ -0,0 +1,44 @@
...
+
+import javax.ejb.EJB;
+import javax.enterprise.context.SessionScoped;
+import javax.faces.bean.ManagedBean;
+
+import org.jboss.as.quickstarts.ssb.MySingletonBean;
+import java.io.Serializable;
+
+@ManagedBean
+@SessionScoped
```
pmuir: This is completely wrong.
1. You are mixing CDI and JSF annotations, which won't work
2. We don't recommend the use of @ManagedBean ever, preferring the use of CDI in all situations.

helloworld-ssb/pom.xml
```diff
@@ -0,0 +1,113 @@
...
+				</configuration>
+			</plugin>
+			<!-- JBoss AS plugin to deploy war -->
+			<plugin>
+				<groupId>org.jboss.as.plugins</groupId>
+				<artifactId>jboss-as-maven-plugin</artifactId>
+				<version>7.1.0.Beta1b</version>
+				<configuration>
+					<username>admin</username>
+					<password>admin</password>
```
pmuir: You shouldn't need to put in a username and password, it should handle just local deployment without it

helloworld-ssb/pom.xml
```diff
@@ -0,0 +1,113 @@
...
+		<dependency>
+			<groupId>org.jboss.spec.javax.annotation</groupId>
+			<artifactId>jboss-annotations-api_1.1_spec</artifactId>
+			<scope>provided</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.jboss.spec.javax.faces</groupId>
+			<artifactId>jboss-jsf-api_2.0_spec</artifactId>
+			<scope>provided</scope>
+		</dependency>
```
pmuir: I don't think we want this dependency in this example.

helloworld-ssb/src/main/java/org/jboss/as/quickstarts/managedbeans/MyManagedBean.java
pmuir: I would recommend just removing the @ManagedBean annotation and adding @Named.

helloworld-ssb/pom.xml
```diff
@@ -0,0 +1,113 @@
...
+	</dependencies>
+
+	<build>
+		<!-- Set the name of the war, used as the context root when the app is 
+			deployed -->
+		<finalName>jboss-as-helloworld-ssb</finalName>
```
pmuir: This must be ${project.artifactId}.

helloworld-ssb/src/main/webapp/WEB-INF/jboss-web.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0"?>
+<!DOCTYPE jboss-web PUBLIC
+   "-//JBoss//DTD Web Application 5.0//EN"
+   "http://www.jboss.org/j2ee/dtd/jboss-web_5_0.dtd">
+<jboss-web>
+	<context-root>/singleton-bean</context-root>
+</jboss-web>	
\ No newline at end of file
```
pmuir: We need to remove this file, the context root must match the artifact id, and this must be controlled in the POM.

helloworld-ssb/src/main/webapp/WEB-INF/web.xml
```diff
@@ -0,0 +1,21 @@
...
+		<url-pattern>*.xhtml</url-pattern>
+	</servlet-mapping>
+	<session-config>
+		<session-timeout>30</session-timeout>
+	</session-config>
+	<welcome-file-list>
+		<welcome-file>home.xhtml</welcome-file>
+	</welcome-file-list>
+</web-app>
\ No newline at end of file
```
pmuir: We should remove this file, we haven't included it in other quickstarts.

pmuir:
Can we do one pull req per quickstart please? :-) It's confusing to review >1 per pull. So remove all the files not related to helloworld-ssb from this one and submit new pulls for those.

helloworld-ssb/src/main/webapp/WEB-INF/jboss-web.xml
spagop: if somebody want to start a project through a quickstart he need to know where the context-root can be set and I think it is important to show that to the users.

pmuir: Then we need a specific quickstart to address that, quickstarts show one concern only.

Also, we will need to work through the tooling implications.

helloworld-ssb/src/main/java/org/jboss/as/quickstarts/managedbeans/MyManagedBean.java
spagop: Good to know what pure CDI means. Already fixed.

helloworld-ssb/pom.xml
spagop: means we need to update other quickstarts.

helloworld-ssb/pom.xml
spagop: already fixed

helloworld-ssb/pom.xml
spagop: thanks for the hint. already fixed.

helloworld-ssb/src/main/webapp/WEB-INF/jboss-web.xml
spagop: already removed.

helloworld-ssb/src/main/webapp/WEB-INF/web.xml
spagop: already removed.

helloworld-ssb/pom.xml
pmuir: Can you let me know which other quickstarts you've seen this in?

spagop: as I see only helloworld-mdb quickstart has not been updated. Can I update and then commit it?

pmuir: Go for it, just send a pull.

sgilda:
I got your changes, tested it, and it works fine in both AS7 and EAP6 now. 

If you don't mind, there are a few minor typos you need to fix before I can merge it:
1. In the README file
   -  You have:
         "start JBoss AS 7.1.0. by running this script". 
     Could you please change that to
          "start JBoss AS 7 (or EAP 6). To do this, run"
   -  You have: 
          "From the home site choose between "Increment A" and "increment B", and then click 
           to the link below you chosen. The link will route you to a corresponding result page. To check if 
           the singleton session bean work fine, from the result page go back to the home page and repeat 
           the same action."
     Could you please change it to something similar to what is on the home page since this where users will 
     look to see if it's what they want. This might be better:
         "This example demonstrates a singleton session bean that maintains state information for 2 variables: 
           "Increment A" and "Increment B". A counter is incremented when you click on the link to the variable name. 
           If you close and restart your browser, or if you have multiple browsers, you can see that the counter 
           always increments the last value. These values are maintained until you restart the server. To test the 
           singleton bean, on the home page, click on either "Increment A" or "Increment B". The result page 
           will give you the current value of the variable."

2.. On the result page, it says:
        Here you're result:
        The character A has been already increment 2 time(s). 

```
Could you change that to:
    Here is your result
    The character A has been incremented 2 time(s).
```

It's looking really good! This should be the last of the changes needed. Thanks!

spagop:
Changes are done and commited. Enjoy.

sgilda:
Thanks Serge! I'll take a look at it when I'm finished with my merge.

sgilda:
Looks good! I'm in the middle of another merge and waiting for help. 
When that's complete, Pete or I will merge your quickstart. Thanks for 
all the hard work!

sgilda:
Pete, I've checked the README, instructions, and page displays. Do you need to check anything else?

pmuir:
I've made a few changes and created them as https://github.com/jbossas/quickstart/pull/43

PR #32, opened 2012-01-03
Quickstart 18  - Programatic control of transactional behaviour

mmusgrov:
Initial commit. Shows how to take control of transaction demarcation and EntityManager enlistment.

### DIFF
--- a/bmt/pom.xml
+++ b/bmt/pom.xml
@@ -1,0 +1,133 @@
+<?xml version="1.0"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+   <modelVersion>4.0.0</modelVersion>
+
+   <groupId>org.jboss.as.quickstarts</groupId>
+   <artifactId>jboss-as-bmt</artifactId>
+   <version>7.0.2-SNAPSHOT</version>
+   <packaging>war</packaging>
+   <name>JBoss AS Quickstarts: Bean Managed Transactions</name>
+   <description>JBoss AS Quickstarts: Bean Managed Transactions</description>
+
+   <url>http://jboss.org/jbossas</url>
+   <licenses>
+      <license>
+         <name>Apache License, Version 2.0</name>
+         <distribution>repo</distribution>
+         <url>http://www.apache.org/licenses/LICENSE-2.0.html</url>
+      </license>
+   </licenses>
+
+   <properties>
+      <!-- Explicitly declaring the source encoding eliminates the following 
+         message: -->
+      <!-- [WARNING] Using platform encoding (UTF-8 actually) to copy filtered 
+         resources, i.e. build is platform dependent! -->
+      <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+   </properties>
+
+   <dependencyManagement>
+      <dependencies>
+         <!-- Define the version of JBoss' Java EE 6 APIs we want to use -->
+         <!-- JBoss distributes a complete set of Java EE 6 APIs including
+            a Bill of Materials (BOM). A BOM specifies the versions of a "stack" (or
+            a collection) of artifacts. We use this here so that we always get the correct
+            versions of artifacts. Here we use the jboss-javaee-6.0 stack (you can
+            read this as the JBoss stack of the Java EE 6 APIs). You can actually
+            use this stack with any version of JBoss AS that implements Java EE 6, not
+            just JBoss AS 7! -->
+         <dependency>
+            <groupId>org.jboss.spec</groupId>
+            <artifactId>jboss-javaee-6.0</artifactId>
+            <version>3.0.0.Beta1</version>
+            <type>pom</type>
+            <scope>import</scope>
+         </dependency>
+      </dependencies>
+   </dependencyManagement>
+
+   <dependencies>
+
+      <!-- Import the CDI API, we use provided scope as the API is included 
+         in JBoss AS 7 -->
+      <dependency>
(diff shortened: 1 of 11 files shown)


bmt/pom.xml
```diff
@@ -0,0 +1,133 @@
...
+         in JBoss AS 7 -->
+      <dependency>
+         <groupId>org.jboss.spec.javax.servlet</groupId>
+         <artifactId>jboss-servlet-api_3.0_spec</artifactId>
+         <scope>provided</scope>
+      </dependency>
+
+      <!-- Import the JPA API, we use provided scope as the API is included 
+          in JBoss AS 7 -->
+      <dependency>
```
pmuir: Formatting has gone a bit crazy here ;-)

bmt/src/main/java/org/jboss/as/quickstarts/bmt/TransactionServlet.java
```diff
@@ -0,0 +1,99 @@
...
+     *
+     * Specify a persistence unit name (perhaps the application may want to interact with multiple
+     * databases).
+     */
+    @PersistenceUnit(unitName = "bmtDatabase")
+    private EntityManagerFactory entityManagerFactory;
+
+    // Inject a UserTransaction for manual transaction demarcation (this object is thread safe)
+    @Resource
+    private UserTransaction userTransaction;
```
pmuir: Not sure if you are aware you can use @Inject here as well if you want.

bmt/src/main/java/org/jboss/as/quickstarts/bmt/ManagedComponent.java
```diff
@@ -0,0 +1,73 @@
...
+     *  Ask the container to inject an Entity Manager (EM). As a consequence the EM will be
+     *  automatically enlisted into any new transactions started by the managed component
+     *  The unit name corresponds to the one defined in the archives persistence.xml file
+     */
+    @PersistenceContext(unitName = "bmtDatabase")
+    private EntityManager entityManager;
+
+    // Inject a UserTransaction for manual transaction demarcation.
+    @Resource
+    private UserTransaction ut;
```
pmuir: You could use @Inject rather than @Resource here. Might be better as it shows a more consistent programming model.

bmt/src/main/java/org/jboss/as/quickstarts/bmt/TransactionServlet.java
```diff
@@ -0,0 +1,99 @@
...
+
+    /*
+     * Inject a stateless bean. Although stateless beans are thread safe it is probably not a
+     * solution that scales particularly well.
+     */
+    @Inject
+    ManagedComponent managedBean;
+
+    // Create a POJO that will manage transactions and the JPA EntityManager itself
+    UnManagedComponent unManagedBean = new UnManagedComponent();
```
pmuir: Rather than instantiate a new object here manually, which will mean that no container services are available in it, including injection, you could use a plain managed bean from CDI which is not transactional anyway (and won't become so in a future EE rev!). This might show a more consistent programming model.

bmt/src/main/java/org/jboss/as/quickstarts/bmt/TransactionServlet.java
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * <p>Servlet entry point.
+     * </p>
+     * <p>The behaviour of the servlet is controlled by servlet query parameters.
+     * If parameters named "key" and "value" are present then that pair is added (or the key is updated if it already
+     * exists) to the database. If the parameter "strategy" is part of the request and has the value "unmanaged" then
+     * both the transaction and the EntityManager are controlled manually. Otherwise the Entity Manager is controlled
+     * by the container and the transaction is controlled by the developer.
```
pmuir, line 69: Personally, I would require "strategy" to be presnent and be one of managed or unmanaged - that just seems clearer to me -- less "magic"...

bmt/src/main/java/org/jboss/as/quickstarts/bmt/ManagedComponent.java
```diff
@@ -0,0 +1,73 @@
...
+         * create an EM for the during of the method call
+         */
+        try {
+            ut.begin();
+
+            /* Since the bean is managed by the container the Entity Manager (EM) and JTA transaction manager (TM) cooperate
+             * so there is no need to tell the EM about the transaction. Compare this with the UnManagedComponent class where
+             * the developer is managing the EM himself and therefore must explicitly tell the EM to join the transaction
+             */
+            String result = UnManagedComponent.updateKeyValueDatabase(entityManager, key, value);
```
pmuir: Not sure that a static method call is the best practice here - see my comments elsewhere about using a CDI managed bean here.

bmt/src/main/java/org/jboss/as/quickstarts/bmt/ManagedComponent.java
```diff
@@ -0,0 +1,73 @@
...
+
+            ut.commit();
+
+            return result;
+        } catch (Exception e) {
+            return e.getMessage();
```
pmuir, line 75: I'm interested that you don't do a rollback here?

pmuir:
A few comments and thoughts on app design, they may not be right - generally really nice and excellent comments!

mmusgrov:
I have accepted all of the above comments. Highlights:
- UnManagedComponent is now a simple CDI bean;
- You said "I would require "strategy" to be present and be one of managed or unmanaged" - I added an HTML checkbox for this parameter

sgilda:
Very nice example! 
Pete, do you want me to attempt the merge?

pmuir:
Pkease

## 

Pete Muir
http://in.relation.to/Bloggers/Pete

On 5 Jan 2012, at 20:30, sgilda
reply@reply.github.com
wrote:

sgilda:
Done. I looked at the commits to jbossas and it appears to be OK.

PR #35, opened 2012-01-04
Quickstart 44

baranowb:
Simple application which logs data to console. Possibly there could be some sort custom config included?


baranowb:
Includes too much. Need to reopen.

PR #36, opened 2012-01-04
Quickstart 44

baranowb:
Simple app to fill console with log4j entries.

### DIFF
--- a/log4jdemo/README.md
+++ b/log4jdemo/README.md
@@ -1,13 +1,76 @@
-Numberguess Example
-===================
+jboss-as-log4jdemo
+========================
 
-This example demonstrates how to use module in  in *JBoss AS 7*.
+What is it?
+-----------
 
-The example can be deployed using Maven from the command line or from Eclipse using
-JBoss Tools.
+This is a simple JSF 2.0 application which will teach you how easy it is to
+use container defined modules to limit size of application package and use 
+common version of certain classes at runitme.
 
-To set up Maven or JBoss Tools in Eclipse, refer to the _Getting Started Guide_.
+It is common for different applications to depend on third party libraries.
+By default, Java EE packages allow to include dependencies in deployable unit.
+This however may lead to uncontrolled growth of deployable unit. To avoid that, 
+one can use container defined module. Module is nothing more than container
+managed binary dependency, which is shared by all deployed applications.
 
-To deploy to JBoss AS 7, start JBoss AS 7, and type `mvn package jboss-as:deploy`. 
-The application is deployed to <http://localhost:8080/jboss-as-log4j>. You
-can read more details in the _Getting Started Guide_.
+For more details on class loading and modules please refer to <https://docs.jboss.org/author/display/AS7/Developer+Guide#DeveloperGuide-JBossDeploymentStructureFile> 
+
+This example is very simple. It declares dependecy on Apache Log4j module to
+allow itself to use custom logging framework.This example achieves this with simple 
+xml file: <b>src/main/webapp/WEB-INF/jboss-deployment-structure.xml</b>.
+For more details about this file please refer to <https://docs.jboss.org/author/display/AS7/Developer+Guide#DeveloperGuide-JBossDeploymentStructureFile>
+
+
+System requirements
+-------------------
+
+All you need to build this project is Java 6.0 (Java SDK 1.6) or better, Maven
+3.0 or better.
+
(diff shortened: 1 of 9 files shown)


sgilda:
Hi Bartosz,

Great example! I have a few comments and suggestions:
1. In the README.md file, could you please make the following changes?
   - It says the URL to run the example is here:
     http://localhost:8080/jboss-as-log4j. I believe it is actually
     here: http://localhost:8080/jboss-as-log4jdemo.
   - The title says "Numberguess Example". Could you change that to
     "Log4j Example"?
   - It says "This example demonstrates how to use module in  in _JBoss
     AS 7_." Could you change that to "This example demonstrates how to
     use the Apache log4j API in your application. It also shows how to
     add a module dependency to the deployment so that these classes
     are visible at runtime."
2. In the home.xhtml, maybe the instructions could say something like: 
   Type a message in the input field and click on Log. You can then view 
   the message in the server console.

Thanks,
Sande

log4jdemo/README.md
```diff
@@ -0,0 +1,13 @@
...
+This example demonstrates how to use module in  in *JBoss AS 7*.
+
+The example can be deployed using Maven from the command line or from Eclipse using
+JBoss Tools.
+
+To set up Maven or JBoss Tools in Eclipse, refer to the _Getting Started Guide_.
+
+To deploy to JBoss AS 7, start JBoss AS 7, and type `mvn package jboss-as:deploy`. 
+The application is deployed to <http://localhost:8080/jboss-as-log4j>. You
+can read more details in the _Getting Started Guide_.
```
pmuir: We generally use a standard layout for the readmes, which confusingly isn't the one followed by numberguess (our bad, we need to fix that!). Take a look at https://github.com/jbossas/quickstart/blob/master/helloworld-jsf/README.md for an example - please can you update to this format.

log4jdemo/pom.xml
```diff
@@ -0,0 +1,128 @@
...
+		<!-- [WARNING] Using platform encoding (UTF-8 actually) to copy filtered 
+			resources, i.e. build is platform dependent! -->
+		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+		<!-- Define version of log4j we want to use. -->
+		<log4j.version>1.2.16</log4j.version>
+	</properties>
+
+	<dependencyManagement>
+		<dependencies>
+			<!-- Define Log4j dependency -->
```
pmuir: Actually, this defines the dependency version, the dependency itself is defined below. Sorry for being pedantic!

log4jdemo/pom.xml
```diff
@@ -0,0 +1,128 @@
...
+			<artifactId>log4j</artifactId>
+			<scope>provided</scope>
+		</dependency>
+
+	</dependencies>
+
+	<build>
+		<!-- Set the name of the war, used as the context root when the app is 
+			deployed -->
+		<finalName>jboss-as-log4jdemo</finalName>
```
pmuir: This should be ${project.artifactId}

log4jdemo/src/main/webapp/WEB-INF/jboss-deployment-structure.xml
```diff
@@ -0,0 +1,9 @@
+<jboss-deployment-structure>
+	<deployment>
+		
+		<dependencies>
+			<!-- Add dependency on log4j, this is not automatic dependency, it has to be explicitly defined, like follows: -->
+			<module name="org.apache.log4j" />
+		</dependencies>
+	</deployment>
+</jboss-deployment-structure>
```
pmuir, line 9: It would be worth calling out this file explicitly in the readme, as it's the key bit of the quickstart!

Mike does an excellent job with readmes, if you want inspiration https://github.com/jbossas/quickstart/pull/32/files#diff-0

pmuir:
Generally looks good, just a few small changes needed to make it excellent!

baranowb:
Updated, with some small tweaks Ive found.

sgilda:
The only change I would suggest is to the following wording, but it certainly isn't necessary. I'm just being picky. :-)
FROM: This example achieves this with simple xml file: : <b>src/main/webapp/WEB-INF/jboss-deployment-structure.xml</b>.
TO: This is achieved by adding the "org.apache.log4j" dependency to the <b>src/main/webapp/WEB-INF/jboss-deployment-structure.xml</b> file.
Pete, are you satisfied with this one now?

pmuir:
Merged

PR #41, opened 2012-01-05
Quickstart 23: Remote access to EJBs

mmusgrov:
This quickstart (authored by Jaikiran) shows how to remotely access stateful and stateless EJBS.
This pull request does not include any other pending quickstarts and It replaces pull request: https://github.com/jbossas/quickstart/pull/33

### DIFF
--- a/dist/src/main/assembly/README.md
+++ b/dist/src/main/assembly/README.md
@@ -25,4 +25,5 @@ contains tutorials for:
 * `numberguess` - CDI + JSF
 * `servlet-async` - CDI + Asynchronous Servlet + Asynchronous EJB 
 * `servlet-filterlistener` - Servlet Filter and Listener
+* `ejb-remote` - shows how to access EJBs remotely (EJB + JNDI)
 
--- a/ejb-remote/README.md
+++ b/ejb-remote/README.md
@@ -1,0 +1,78 @@
+Remote EJB Client Example
+=========================
+
+What is it?
+-----------
+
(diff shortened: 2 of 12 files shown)


pmuir:
Sande, I'm happy with this, so once you are, we can merge it.

sgilda:
Hi Mike,
This works fine against jboss-as-7.1.0.Final-SNAPSHOT, but not against jboss-eap-6.0.0.DR9. I'm guessing fixes were needed after DR9 for this example.
Looks great!
Pete, I attempted the merge but ran into a problem on this step: git merge ejb-remote. I'll send an email.

mmusgrov:
It is probably an issue with the version of jboss-ejb-client:

On jboss-as-7.1.0.Beta1b I was using version Beta7 of jboss-ejb-client
But on jboss-as-7.1.0.Final-SNAPSHOT only Beta9 works.

My guess is that jboss-eap-6.0.0.DR9 hasn't got the Beta9 version yet.

This is a problem for the client part of the example since we cannot use <scope>provided</cope> in the maven pom.

Any thoughts on how to manage the client dependencies (there are other dependencies in the client pom that are also still at beta revs).

sgilda:
This is beyond me. Pete?

pmuir:
It's just a versioning problem - we are using a different version in EAP vs AS 7. This is a problem as the Java EE BOM doesn't include this jar, so we can't let it manage the dependency for us.

I've contacted Shelly and Paul to find out what the plans are here.

PR #46, opened 2012-01-09
Corrected the startup instructions for the MDB example

tomjenkinson:
I _may_ be wrong but I couldn't get the helloworld-mdb example to work on my machine. The issue was the the connection factory and queue do not exist in standalone.xml, rather standalone-full.xml, hence my patch (I also didn't need to create the admin user so removed that comment also).

I realise that there isn't a Jira number for this so feel free to reject the pull req. Lets say you do reject it, do we raise Jiras in the main AS7 project for the quickstarts also?

Sorry for the unorthodox method of doing this, but I wasn't sure how we were tackling discovering issues, my main entry into this work being: https://docs.jboss.org/author/display/AS71/Quickstarts

Ta,
Tom

### DIFF
--- a/helloworld-mdb/README.md
+++ b/helloworld-mdb/README.md
@@ -23,13 +23,13 @@ With the prerequisites out of the way, you're ready to build and deploy.
 Deploying the application
 -------------------------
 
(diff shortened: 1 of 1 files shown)


pmuir:
Thanks tom, I merged this.

We aren't really tracking issues properly at the moment, we've found pull requests work well. You can use AS7 as well if you want.

PR #51, opened 2012-01-10
fixed declaration of JMS state to not be static

tomjenkinson:
Seemed strange that the JMS static was statically declared


pmuir:
Merged

PR #68, opened 2012-01-20
JTS example

tomjenkinson:
Although not listed on: https://docs.jboss.org/author/display/AS71/Quickstarts I thought that providing a JTS example would be useful as it is a fairly complex deployment scenario.

If you do accept the quickstart, should one of us update https://docs.jboss.org/author/display/AS71/Quickstarts with the additional quickstart?

The example requires two instances of an AS, it deploys an EJB into one of them and into the other it deploys another EJB which uses it and a small JSF application.
Each EJB will publish a message locally (using XA) and a dummy MDB reads off the message (clearly once the distributed transaction has committed).

Let me know what you think!


pmuir:
Sande, could you do an initial review.

sgilda:
Sure.

sgilda:
I love this quickstart! The code is clear and easy to understand. You also did a great job with the README instructions and explaining how to set it up and how it works.

I do have a some suggestions for the README, but they are by no means necessary.  :-)

It would be nice if the first paragraph had a little more information and maybe spelled out some of the acronyms. Maybe something like this: 
    This example demonstrates how to perform distributed transactions in an application. 
    It uses Java Transaction Service (JTS) to propagate a transaction context across 
    two Container-Managed Transaction (CMT) EJBs that, although deployed in separate servers,
    participate in the same transaction. In this example, one server processes the 
    Customer and Account data and the other server processes the Invoice data.

Maybe add a separate section "Configuring and Starting the Servers" before "Testing the Application" and moving all the server configuration there? Since this example requires a little more setup than usual, it might be nice to make it stand out.

I love the way you use the replaceable values for APP_SERVER_1_HOME and other things. That might be a good approach for the server configuration file instructions too. Something like this (this is very rough):

```
Due to differences in configuration, you will need to to specify a different configuration file 
when you start the server, depending on whether you are running AS7 or EAP 6. 

For AS 7, you need to specify standalone-full.xml for the <SERVER_CONFIGURATION_FILE>.
For EAP 6, you can specify standalone.xml for <SERVER_CONFIGURATION_FILE>, or omit it 
from the command line since it is the default.
```

Then, later in the text use the <SERVER_CONFIGURATION_FILE> in place of "standalone-full.xml'.

I'll email a mockup of what I'm talking about so you can see whether you want to make the changes or leave it as is.

sgilda:
Tom, I just noticed some of the pom.xml versions are incorrect. Since Pete wrote scripts to modify version numbers when quickstarts are moved from AS7 to EAP6, they need to match the AS 7 versions specified here, otherwise the scripts don't work well: https://docs.jboss.org/author/display/AS71/Maven+POM+Versions+Checklist
For example:
Tthe org.jboss.as.quickstarts (groupId) version needs to be changed from 7.1.0.Final-SNAPSHOT to 7.0.2-SNAPSHOT in all of your pom.xml files.
The maven-compiler-plugin is 2.3.2 instead and should be 2.3.1.
The jboss-as-maven-plugin is 7.1.0.Beta1b  and should be 7.1.0.CR1.
I didn't check everything, but you get the idea. :-)

sgilda:
Tom, I just looked at the new wording for your cmt README changes for starting with a full profile. I like your wording much better than mine! Your from cmt (below) is very clear! Please use that instead of mine. :-)
For AS7:
        $JBOSS_HOME/bin/standalone.sh -c standalone-full.xml
or if you are using AS7 on windows
        $JBOSS_HOME/bin/standalone.bat -c standalone-full.xml

For EAP6:
        $JBOSS_HOME/bin/standalone.sh
or if you are using EAP6 on windows
        $JBOSS_HOME/bin/standalone.ba

sgilda:
This looks great to me. I like the way you went even further in the README and added details about ACID transaction properties! It's a term developers might search on when looking for an example. The POMs look good too.
Pete, if you don't have any other suggestions or changes, I think this one is good to go.

sgilda:
Tom, at some point in the future, would it be possible to put some sort of message to the page saying the transaction was successful? Your README explains you have to look in the server logs, so I don't think it needs to be done now, but eventually it would be nice to put feedback on the page saying the transaction was successful and to look in the first server log for the message showing the customer was created and the second server log for the message saying the invoice was created. Maybe sometime before final, if it's possible?

sgilda:
Hi Tom,
That was fast! 
I know I am being really picky here, but is it possible to have the results display above or below the input data on the same page, so you don't have to navigate back to the original page to add another invoice?  
Also, I don't think you need to put that much detail. My first reaction was that I had an error. Something like "Transaction succeeded. For detailed information, see the 'Created customer' message in the log for server 1 and the 'Created invoice' message on server 2." would be sufficient.
You may want to change the last line in the README that says the page will not change since it will now display a message :-).
Thanks,
Sande

sgilda:
Very nice! Thanks for being so patient with me.
Pete, if you are OK with this one, I think it's good to go.

jts/README.md
```diff
@@ -0,0 +1,123 @@
...
+* Consistent
+* Isolated
+* Durable.
+
+The example uses Java Transaction Service (JTS) to propagate a transaction context across 
+two Container-Managed Transaction (CMT) EJBs that, although deployed in separate servers,
+participate in the same transaction. In this example, one server processes the 
+Customer and Account data and the other server processes the Invoice data.
+
+The example expects to be deployed onto the same physical machine. This is not a restriction
```
pmuir: We could perhaps say "The example expects >>>the EJBs<<< to be deployed onto the same physical machine" as we've had a new sentence since the subject was introduced.

jts/README.md
```diff
@@ -0,0 +1,123 @@
...
+of JTS and the example can easily be converted to run on separate machines by editing the
+hostname value for EJB2 in org.jboss.as.quickstarts.cmt.jts.ejb.AccountManagerEJBImpl.
```
pmuir: Could we externalise this somehow? It seems bad to suggest people hard coding this. I realise there is a dichotomy between simplicity and correctness here. Just wondering ;-)

jts/application-component-1/pom.xml
```diff
@@ -0,0 +1,138 @@
...
+		<!-- Import the EJB API, we use provided scope as the API is included in 
+			JBoss AS 7 -->
+		<dependency>
+			<groupId>org.jboss.spec.javax.ejb</groupId>
+			<artifactId>jboss-ejb-api_3.1_spec</artifactId>
+			<scope>provided</scope>
+		</dependency>
+		<!-- Import the injection annotations -->
+		<dependency>
+			<groupId>javax.inject</groupId>
```
pmuir: Normally here we would expect people to use the CDI api which has a transitive dependency on @Inject. You are actually using CDI in the example I suspect, but only a subset of it. That will remove the need to add the version explicitly.

pmuir:
I'm wondering why we've started using EJBHome here, rather than the EJB 3 programming model. This should be explained in the README if it's really necessary.

jts/README.md
tomjenkinson: Done in commit

jts/application-component-1/pom.xml
tomjenkinson: Done in commit, though I usually refrain from pulling in a dep just to pull in another

pmuir:
> Ok, so the only things I think are remaining are:
> 1. EJB home for remote ejb - am I correct I have to use this? I believe so but will change it if you link me to something

I've asked Jaikiran to take a look.

> 1. Externalising the word localhost from application-component-1/src/main/java/org/jboss/as/quickstarts/cmt/jts/ejb/AccountManagerEJB.java line 37 - I am happy to do this, but can you recommend a pattern for the quickstarts. I happen to feel that asking the user to change that one reference isn't too onerous but will make the change if you like.

I'm not sure of the best way to do it tbh. Let's keep it in the back of our minds that if we need to do this a lot, we need a general solution.

jaikiran:
> Re: "I'm wondering why we've started using EJBHome here, rather than the EJB 3 programming model. This should 
> be explained in the README if it's really necessary." This is legacy from my earlier stab at this, have removed the 
> EJBhome where possible, but I think a remote EJB needs to be as I get a classcastexception resolving, e.g. 
> corbaname:iiop:localhost:3628#jboss-as-jts-application-component-2/InvoiceManagerEJBImpl as an instance, it is a 
> home so I think I have to do this?

For IIOP invocations, we do mandate that the EJB expose a EJB2.x remote home view and that's the view we use for allowing the invocations. So yeah, Tom will need a EJB2.x style home interface exposed for that bean for a IIOP invocation.

pmuir:
Sande, this is ready for merge then.

sgilda:
Done!

PR #70, opened 2012-01-20
Helloworld mdb

sgilda:
Modified the README.md instructions for starting the server so that it works for both AS7 and EAP6.


pgier:
Merged a single squashed commit in ac12124076ebd40d68cd1d1c956b2963561d55e9

PR #92, opened 2012-02-06
Jax rs client

pmuir:


sgilda:
I noticed the following differences in the versions specified in the pom.xml file compared to the Maven POM Versions Checklist
-- maven-compiler-plugin artifactId: 2.3.2. v.s. 2.3.1 in the Checklist
-- junit artifactId: 3.8.1 v.s. 4.10 in the Checklist

In the README file, we should refer to other quickstarts by the quickstart name, not the number. When developers download them, there is no reference to the number so they won't have a clue which one it is. I had to go to the sign-up sheet to find out that quickstart #32 is helloworld-rs. :-)

Also, the fails with a 404 Not Found error because the rs-helloworld quickstart was renamed to helloworld-rs. The URLs need to change in this quickstart in the README file and in the JaxRsClient.java file.

pmuir:
Blaine has now submitted again.

PR #98, opened 2012-02-10
Jax rs client

blainemincey:
Updated README.md with correct references to helloworld-rs.  Corrected versions in pom.xml for JUnit and maven compiler.  Corrected helloworld-rs URL in src and added additional comments.


pmuir:
My only comments are that:
- the java source file could do with some comments
- Rather than using the maven exec command, why not make it a real JUnit test? This would handle a lot of the boilerplate you have in the POM and in your class

Sande, WDYT?

blainemincey:
Hi Pete and Sande.

Added additional comments to java src.  Moved the java src to use the JUnit4 test harness.  Made the suggested changes to the POM.

pmuir:
Looks really good now. I'll let Sande do a review too, but I'm happy.

sgilda:
Looking now.

sgilda:
Hi Blaine,
This is easy to follow and works great! 

I have a few minor suggestions for the README.md file:
-- Remove the hyphen from "Pre-requisite" and add an "s": "Prerequisites".
-- You say: This example depends on quickstart #32 being built and deployed to AS7. I don't think many developers will kno w the number sequences of the quickstarts since they aren't referenced in the download or the README files. Could you change that to "Before running this quickstart, you must build and deploy the helloworld-rs quickstart to your AS 7 server as this quickstart depends on that deployment for its test." (or something like that)
Later, when you show the 2 URLS, could you put them on separate lines?  It's hard to tell there are 2 when they're on the one line like this: http://localhost:8080/jboss-as-helloworld-rs/xml http://localhost:8080/jboss-as-helloworld-rs/json

Other than that, this is good to go!

sgilda:
Pushed upstream.

sgilda:
Closed pull request.

PR #101, opened 2012-02-12
Use transactions and tools boms

paulrobinson:
Updated to use the jboss-javaee-6.0-with-tools and jboss-javaee-6.0-with-transactions boms


pmuir:
Merged

PR #110, opened 2012-02-16
[AS7-3634] revisited: sync up with AS 7.1.0.Final -- these will be compa...

rhusar:
...tible with future versions


sgilda:
This fixes it. The ejb-remote quickstart works great now.

PR #119, opened 2012-02-27
Eap pom module fix

sgilda:
I modified the pom.xml file to add the missing quickstart modules. I also added the jax-rs-client to the README files in the root and in the dist directory.


sgilda:
This pull was issued to the wrong repository. Will close and reopen.

PR #121, opened 2012-02-28
kitchensink: facelet cleanup

bleathem:
Replaced the html table markup with a JSF h:panelGrid
Changed the success message to use FacesMessages

### DIFF
--- a/kitchensink/pom.xml
+++ b/kitchensink/pom.xml
@@ -104,6 +104,14 @@
          </exclusions>
       </dependency>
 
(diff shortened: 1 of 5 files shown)


sgilda:
Pete, do we also need to modify the archetype that this is based on? 

I think this will also impact the other quickstarts that were based on this one. So far it looks like the following might need to be updated:

--  helloworld-singleton (not sure this one needs to be modified)
--  hibernate3
--  hibernate4
--  kitchensink-ear + its archetype
--  kitchensink archetype

Pete, if you are good with this one, I can push it.

PR #125, opened 2012-02-29
kitchensink: Refactored the database manipulation into a service layer to facilitate testing

bleathem:


sgilda:
There are 2 updates to the kitchensink quickstart. This one and https://github.com/jbossas/quickstart/pull/126. 
Are they going to trip over each other?

bleathem:
Yes, they will collide, if the approach I'm suggesting here is acceptable, pull/126 will have to be refactored

sgilda:
OK. I will test this one first then.
Who needs to review this change? 
Do we need to modify the archetype that this quickstart is generated from also?

bleathem:
This pull request involves a design change for the approach taken with the quickstarts so far, by introducing a service layer, and removing model concerns from the controller.  Whoever is the gate keeper for such changes would have to review this pull request (Pete I'm guessing?)  IMO the introduction of the service layer makes it cleaner to integrate with RESTful service endpoints.

This change will have to be propagated to the archetypes, and other quickstarts.

sgilda:
I tested this and it works fine.
Who needs to do the code review?

pmuir:
Merged. I am SME for this :-)

PR #126, opened 2012-02-29
Kitchensink upgrade

pmuir:


kitchensink/src/main/java/org/jboss/as/quickstarts/kitchensink/controller/CountriesController.java
```diff
@@ -0,0 +1,136 @@
...
+   @Produces @Named
+   public Country getCountry() {
+      return country;
+   }
+   
+   public void add() throws Exception {
+      if (code == null) {
+         log.info("Registering " + country.getName());
+         em.persist(country);
+         FacesContext.getCurrentInstance().addMessage(null, new FacesMessage(SEVERITY_INFO, "Added " + country.getCode() , null));
```
bleathem, line 93: I'd put a FacesContest producer in the org.jboss.as.quickstarts.kitchensink.Resources so we can @Inject the FacesContext, rather than use the static lookup.

bleathem:
Looks good (when I mentally project my earlier changes onto it) - only thing I'd suggest is that we create some composite components to mask the complexity of the index page.  I did this in kitchensink-rf by creating a composite component for the member form elements.  That being said, such a change can be added after merging this request.

pmuir:
Nice idea.

sgilda:
There are 2 updates to the kitchensink quickstart. This one and https://github.com/jbossas/quickstart/pull/125. 
Are they going to trip over each other?

pmuir:
Yes. I will have to merge them ;-)

kitchensink/src/main/java/org/jboss/as/quickstarts/kitchensink/controller/CountriesController.java
```diff
@@ -0,0 +1,136 @@
...
+   
+   @Inject
+   private Event<Country> countryEventSrc;
+   
+   private List<Country> countries;
+
+   private Country country;
+   private String code;
+
+   @Produces @Named
```
StevenBoscarine, line 58: Could this method use a JavaDoc?  Would it help out the user to explain what we're doing and why?

kitchensink/src/main/java/org/jboss/as/quickstarts/kitchensink/data/CountryListProducer.java
```diff
@@ -0,0 +1,39 @@
...
+import javax.enterprise.event.Observes;
+import javax.enterprise.event.Reception;
+import javax.enterprise.inject.Produces;
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.persistence.EntityManager;
+
+import org.jboss.as.quickstarts.kitchensink.model.Country;
+
+@RequestScoped
```
StevenBoscarine, line 16: Do you want a class level comment here?

kitchensink/src/main/java/org/jboss/as/quickstarts/kitchensink/controller/CountriesController.java
```diff
@@ -0,0 +1,136 @@
...
+   public String getCode() {
+      return code;
+   }
+   
+   public void setCode(String code) {
+      this.code = code;
+   }
+   
+   public void retrieve() {
+      if (code != null) {
```
StevenBoscarine, line 119: What's the conversation used for?  Why are we checking if it's transient? Would it help out the user to explain what we're doing and why?

kitchensink/src/main/java/org/jboss/as/quickstarts/kitchensink/controller/MembersController.java
```diff
@@ -0,0 +1,121 @@
...
+   }
+   
+   public String getEmail() {
+      return email;
+   }
+   
+   public void setEmail(String id) {
+      this.email = id;
+   }
+
```
StevenBoscarine, line 82: Could this method use a JavaDoc? Would it help out the user to explain what we're doing and why?

kitchensink/src/main/java/org/jboss/as/quickstarts/kitchensink/controller/MembersController.java
```diff
@@ -0,0 +1,121 @@
...
+            conversation.begin();
+         }
+         List<Member> members = em.createQuery("select m from Member m where m.email = :email").setParameter("email", email).getResultList();
+         if (members.isEmpty()) {
+            // This could happen if the user manually sets the wrong email addr
+            FacesContext.getCurrentInstance().addMessage("member", new FacesMessage(SEVERITY_ERROR, "Unable to load member with email " + email, null));
+         } else if (members.size() > 1) {
+            // This should never happen, it means a db constraint went wrong
+            throw new IllegalStateException("Something went very wrong, email should be unique!");
+         } else {
```
StevenBoscarine, line 95: What's going on here?  Would it help out the user to explain what we're doing and why?

kitchensink/src/main/java/org/jboss/as/quickstarts/kitchensink/model/Address.java
```diff
@@ -0,0 +1,210 @@
...
+      if (region == null) {
+         if (other.region != null)
+            return false;
+      } else if (!region.equals(other.region))
+         return false;
+      return true;
+   }
+
+   public String toSimpleString() {
+      StringBuilder builder = new StringBuilder().append(address1);
```
StevenBoscarine, line 198: why not StringBuilder builder = new StringBuilder(address1);  ?

kitchensink/src/main/java/org/jboss/as/quickstarts/kitchensink/model/Member.java
```diff
@@ -11,44 +13,98 @@
...
    @Id
-   @GeneratedValue
-   private Long id;
-
-   @NotNull
-   @Size(min = 1, max = 25)
+   @GeneratedValue(strategy = IDENTITY)
+   // Tell JAXB to ignore the synthetic ID when generating an XML model. It's not needed
+   @XmlTransient
+   private long id;
```
mbogoevici, line 51: Any particular reason for the Long->long switch? I think that the ID should be remain an Object (thus allowing for null in transient instances)  rather than a primitive.

kitchensink/src/main/java/org/jboss/as/quickstarts/kitchensink/model/Member.java
```diff
@@ -75,4 +131,40 @@ public String getPhoneNumber() {
    public void setPhoneNumber(String phoneNumber) {
       this.phoneNumber = phoneNumber;
    }
+
+   public Address getAddress() {
```
mbogoevici, line 135: Mark with @Valid, lest validation will not be cascaded.

sgilda:
Are we still waitng for changes to this one?

pmuir:
Right now we don't have agreement about whether to include this pull in the quickstarts or not.

pmuir:
Thanks to Steven and Marius for their comments. I will address once we decide what to do with this pull!

pmuir:
Pull defered until after EAP 6.

PR #130, opened 2012-03-01
Added dependency to pom.xml for httpcore

blainemincey:
When testing with the JBoss provided Maven repository, an additional dependency was required for this quickstart.  Specifically, the following was added to the POM:

<dependency>
    <groupId>org.apache.httpcomponents</groupId>
    <artifactId>httpcore</artifactId>
    <version>4.1.4</version>
 </dependency>


sgilda:
Thanks Blaine. Pushed and closed.

PR #136, opened 2012-03-08
Changed to using own queues rather than testQueue

tomjenkinson:


sgilda:
Merged and closed. Thanks Tom!

PR #139, opened 2012-03-10
Merge 1.0.0.M2b from aerogear/quickstarts

kborchers:


kborchers:
We'll need to make some more updates to this and do an M2c so closing for now.

PR #141, opened 2012-03-12
change to use jboss-javaee-6.0-with-tools bom

stliu:


hibernate4/pom.xml
```diff
@@ -129,10 +125,21 @@
...
+           <scope>provided</scope>
+       </dependency>
+
+       <!-- use for JPA Meta model generation, used only in development time, so set it to provided scope -->
+       <dependency>
+           <groupId>org.hibernate</groupId>
+           <artifactId>hibernate-jpamodelgen</artifactId>
+           <scope>provided</scope>
+       </dependency>
       <!-- Needed for running tests (you may also use TestNG) -->
```
pmuir, line 143: Do we need to add the above? Are they used?

log4j/pom.xml
```diff
@@ -47,9 +49,9 @@
 				the JBoss stack of the Java EE 6 APIs). You can actually use this stack with 
 				any version of JBoss AS that implements Java EE 6, not just JBoss AS 7! -->
 			<dependency>
-				<groupId>org.jboss.spec</groupId>
-				<artifactId>jboss-javaee-6.0</artifactId>
-				<version>3.0.0.Beta1</version>
+                <groupId>org.jboss.bom</groupId>
+                <artifactId>jboss-javaee-6.0-with-tools</artifactId>
+                <version>${javaee6.with.tools.version}</version>
```
pmuir: Do you need to remove some dependencies elsewhere, otherwise, why the change?

hibernate4/pom.xml
stliu, line 143: they are not used _yet_, since this example project is not finished, and as a example project of hibernate4, I would like it shows some key features from JPA2 --  the type safe criteria query, powered by _hibernate-jpamodelgen_

log4j/pom.xml
stliu: sorry, i'm confusing about these two projects, whats the different between _org.jboss.spec:jboss-javaee-6.0_ and _org.jboss.bom:jboss-javaee-6.0_?

pmuir: Assuming you mean " whats the different between `org.jboss.spec:jboss-javaee-6.0` and `org.jboss.bom:jboss-javaee-6.0-with-tools`", then the latter includes Arquillian, whilst the former is just Java EE.

hibernate4/pom.xml
pmuir, line 143: I suggest we only add them as needed, as otherwise intermediate releases of the quickstart are wrong. This quickstart is included in the Quickstart set today.

pmuir:
Strong, just a reminder we are waiting on your changes here.

stliu:
@pmuir I have updted my commit as your pointed out, sorry I must missed the previous mail notification

pmuir:
I fixed the various formatting, naming convention problems etc. and merged. Take a look at what i changed.

PR #143, opened 2012-03-13
Updated quickstart to work with new JNDI implementation and security in ...

jbertram:
...AS 7.1.0.Final

### DIFF
--- a/helloworld-jms/README.md
+++ b/helloworld-jms/README.md
@@ -16,25 +16,19 @@ The example provides for a certain amount of customization for the `mvn:exec` pl
 Required Properties
 -------------------
 
-* `connection.type`
+* `username`
    
-   Which connection type to the JMS provider you want to use. Accepted values are `netty` or `jndi`.
+   This username is used for both the JMS connection and the JNDI look-up (if necessary).  Use the bin/add-user.sh script on the server
+   to add this user to the "guest" role and the "ApplicationRealm".
    
-   Default: `netty`
-   
-   *Note:* Currently external JNDI access is under development in JBoss AS 7. As a result netty is the default value.
+   Default: `defaultUser`
 		
-* `java.naming.factory.initial`
+* `password`
 
-   The JNDI initial context factory you wish to use
+   This password is used for both the JMS connection and the JNDI look-up (if necessary).  See note about "username" regarding how to add
+   the proper credentials to the server.
    
-   Default: `org.jboss.as.naming.InitialContextFactory`
-
-* `java.naming.provider.url`
-   
-   The provider URL of the JNDI context you wish to use
-
-   Default: `jnp://localhost:1099`
+   Default: `defaultPassword`
 
 
 Optional Properties
@@ -42,13 +36,13 @@ Optional Properties
 
 * `cf.name`
 
-   The JNDI name of the JMS ConnectionFactory you want to use.
+   The name of the JMS ConnectionFactory you want to use.
 
    Default: `RemoteConnectionFactory`
 
 * `dest.name`
 
-   The JNDI name of the JMS Destination you want to use.
+   The name of the JMS Destination you want to use.
    
    Default: `testQueue`
 
--- a/helloworld-jms/pom.xml
+++ b/helloworld-jms/pom.xml
@@ -36,12 +36,31 @@
          <groupId>org.jboss.spec.javax.jms</groupId>
          <artifactId>jboss-jms-api_1.1_spec</artifactId>
          <version>1.0.0.Final</version>
-         <scope>provided</scope>
       </dependency>
       <dependency>
          <groupId>org.jboss.as</groupId>
          <artifactId>jboss-as-messaging</artifactId>
          <version>7.1.0.Final</version>
+      </dependency>
+      <dependency>
+         <groupId>org.jboss.marshalling</groupId>
+         <artifactId>jboss-marshalling-river</artifactId>
+         <version>1.3.9.GA</version>
+      </dependency>
+      <dependency>
+         <groupId>org.jboss.xnio</groupId>
+         <artifactId>xnio-api</artifactId>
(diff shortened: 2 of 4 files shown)


sgilda:
Hi Justin,
Apparently there is some additional setup required for this quickstart? I'm getting this error:

SEVERE: Unable to validate user: myUser
javax.jms.JMSSecurityException: Unable to validate user: myUser

I found this in the pom.xml file:



```
              <systemProperty>
                 <key>username</key>
                 <value>myUser</value>
              </systemProperty>
              <systemProperty>
                 <key>password</key>
                 <value>myPass</value>
```

I added a myUser/myPass User with a guest role to the application realm, but now I am seeing this error:

SEVERE: User: myUser doesn't have permission='CONSUME' on address jms.queue.testQueue
javax.jms.JMSSecurityException: User: myUser doesn't have permission='CONSUME' on address jms.queue.testQueue

The README instructions need to be updated. I'll email you a copy so you can verify that I set it up correctly.

Any idea what I did wrong?

Do we have to use the bin/add-user.sh script to add a user? To which 
Can you modify the README file

jbertram:
I did update the README.md.  It's part of the commit.  Did you see those changes?

jbertram:
Make sure your user is in the "guest" role as stated in the README.md and in the pom.xml.

sgilda:
I was looking at the wrong version of the README. Sorry! My bad. :-)

sgilda:
I'm pretty sure I did.

sgilda:
In fact, I copied and pasted from the command line for the readme. Maybe there was an extra space somewhere.

Is there a way to remove a user so I can add them back?

jbertram:
Are you using AS 7.1.0.Final or the latest snapshot of 7.1.1?  AS 7.1.0.Final has a bug that may prevent add-user.sh from working properly.  I mentioned this in my email to you earlier today in a reply to the thread you forwarded to me.  I recommend you use 7.1.1 because of this.

sgilda:
I'm using AS 7.1.0.Final. That must be it. Thanks!

jbertram:
Once you try with 7.1.1 let me know if you have any issues.  Thanks!

Also, I think the walk-through you added for using add-user.sh is worth keeping.

pmuir:
I wonder if it's worth removing the whole netty/jndi option thing as part of this? I assume JNDI is recommended, and as this is a quickstart, it should show recommended approaches not all. I think this could simplify the code considerably.

Sande, we should add the stuff about adding users to the root readme as a number of quickstarts will need this.

sgilda:
It does make sense. It also would be nice if all the quickstarts were 
modified to use the same user. Maybe quickstartUser/quickstartPassword 
or something like that.

pmuir:
Good idea.

sgilda:
I can't download it at the moment because the web site is down for 
maintenance. I assume it's available here: 
http://www.jboss.org/jbossas/downloads/

I'll try again in an hour.

sgilda:
Hi Justin,

It does work great with JBoss 7.1.1. Thanks!

I am going through the README as a new user and I think the instructions need to be more specific. For example, it states:

```
    * `username`

        This username is used for both the JMS connection and the JNDI look-up (if necessary).  Use the bin/add-user.sh script on the server to add this user to the "guest" role and the "ApplicationRealm".

         Default: `defaultUser`
```

When I read that as a new user, I assume I create a user "defaultUser/defaultPassword" or "username/password". Nowhere does it indicate that the username you need to create is "myUser/myPass".

It's not until you compile that you know the username you need to create:

```
   SEVERE: Unable to validate user: myUser
   javax.jms.JMSSecurityException: Unable to validate user: myUser
```

Also, why does it show BUILD SUCCESS?

```
    Caused by: HornetQException[errorCode=105 message=Unable to validate user: myUser]
        ... 13 more
     [INFO] ------------------------------------------------------------------------ 
     [INFO] BUILD SUCCESS
     [INFO] ------------------------------------------------------------------------
```

Would you mind making the following changes?

--  Change the username/password from myUser/myPass to quickstartUser/quickstartPassword in anticipation of trying to have one user for all the quickstart.
--  Modify the README as follows:
---- Move the instructions to start the server before the maven compile instructions. I'm afraid some users may follow the steps and not realize until the end that the server must be started
---- Add specific details about the user to create. You might want to point out the properities in the pom.xml file and how they impact this. Eventually I'll pull the add-user instructions to the root readme, but the pom.xml information needs to be here.
---- Add information about what you should expect to see when you  successfully run the quickstart. For example:
Mar 14, 2012 8:21:47 AM org.jboss.as.quickstarts.jms.JMSClientUtil getConnectionFactory
INFO: Attempting to acquire ConnectionFactory "RemoteConnectionFactory" with a connection type of: netty
Mar 14, 2012 8:21:47 AM org.jboss.as.quickstarts.jms.JMSClientUtil getConnectionFactory
INFO: Creating Netty Based ConnectionFactory.
Mar 14, 2012 8:21:47 AM org.jboss.as.quickstarts.jms.JMSClientUtil getDestination
INFO: Attempting to acquire Destination with a connection type of: netty
Mar 14, 2012 8:21:47 AM org.jboss.as.quickstarts.jms.JMSClientUtil getDestination
INFO: HornetQ Destination testQueue
Mar 14, 2012 8:21:48 AM org.jboss.as.quickstarts.jms.HelloWorldJMSClient main
INFO: Sending 1 messages with content: Hello, World!
Mar 14, 2012 8:21:48 AM org.jboss.as.quickstarts.jms.HelloWorldJMSClient main
INFO: Received message with content Hello, World!

Thanks for getting this quickstart working again!
Sande

jbertram:
Yes, I can do that.  Do you want me to make this a JNDI only quick-start?  If we want to show the HornetQ "core" API (i.e. Netty) use-case, we could have a different quick-start for that.

pmuir:
+1

PR #150, opened 2012-03-19
AS7-4063 logging tools quick start

dmison:
Quick start that uses a simple jax-rs service to demo using jboss-logging-tools for i18n/l10n of exceptions, logging and generic strings.


sgilda:
Hi Darrin,
Very nice quickstart! I like the fact that it's localized too. It's our first localized quickstart.
I do have a few comments and suggestions:
- The pom.xml file has a few incorrect versions:
  -- org.jboss.as.quickstarts (groupId): change version from 7.0.2-SNAPSHOT 7.1.1-SNAPSHOT
  -- maven-compiler-plugin: you may need to change the version from 2.3.2 to 2.3.1. I need to check on this because I'm not really sure. The others all have 2.3.1.
- Could you rename the quickstart to logging-tools. There is no need for the -qs appended at the end
- The URL should be standardized with the 'jboss-as-" prefix: http://localhost:8080/jboss-as-logging-tools. This is to make it unique and prevent it from overwriting an existing URL on the server. This wasn't in the quickstart instructions, so I added it. 
- Could you change the title from "i18n/l10n with JBoss Logging Tools" to "Internationalization and Localization (i18n/l10n) with JBoss Logging Tools". We try to spell out the acronyms first.
- In the README file, could you tell them to back up the standalone.conf file before adding the optional JAVA_OPTS? I know it's a simple change and seems silly, but in case of mistakes, it's better to be safe than sorry.  Also remind them to restore the backup version of the file when done testing.
- I believe you can add just the one line like this: JAVA_OPTS="$JAVA_OPTS -Duser.country=DE -Duser.language=de"
- I find it confusing to use the greetings URL with the name "crashme" for the exception. Is it possible to change the URL ttp://localhost:8080/jboss-as-logging-tools/rest/error or http://localhost:8080/jboss-as-logging-tools/rest/error/divide-by-zero to make it unique from the greetings URL?  I see you are trying to show that the specific path for name "crashme" does something different, but it just seems odd to me for this example.
- When you put out the response, for example "Hello Harold" or the number of days, would it be possible to add a note telling them how to change the URL to get a different response? It's not mandatory, but nice to have.

Pete, who should do the code review for this one?

Thanks, Sande

pmuir:
James Perkins would be good.

logging-tools-qs/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+         $JBOSS_HOME/bin/standalone.sh
+
+2. Build the application
+
+         mvn clean package
+	   
+3. Deploy the application 
+
+         mvn jboss-as:deploy
```
pmuir: The markdown formatting of this list is wrong - take a look in any markdown renderer, or run dist/release-utils.sh -m and then try to fix it.

logging-tools-qs/README.md
```diff
@@ -0,0 +1,116 @@
...
+	   
+## Running the quick start
+
+Once deployed, you can access the quick start using the following URLs.
+
+### http://localhost:8080/logging-tools-qs/rest/greetings/`name`
+
+   Demonstrates simple use of localised messages (with parameter) and logging.
+
+   Example: [http://localhost:8080/logging-tools-qs/rest/greetings/Harold](http://localhost:8080/logging-tools-qs/rest/greetings/Harold)
```
pmuir: If you < and > here you can get a much simpler link syntax :-)

logging-tools-qs/pom.xml
```diff
@@ -0,0 +1,128 @@
...
+         in JBoss AS 7 -->
+      <dependency>
+         <groupId>org.jboss.spec.javax.ws.rs</groupId>
+         <artifactId>jboss-jaxrs-api_1.1_spec</artifactId>
+         <scope>provided</scope>
+      </dependency>
+      <dependency>
+      	<groupId>org.jboss.logging</groupId>
+      	<artifactId>jboss-logging-processor</artifactId>
+      	<version>1.0.0.Final</version>
```
jamezp: This will need to be updated to 1.0.1.Final because of a translation bug.

logging-tools-qs/pom.xml
```diff
@@ -0,0 +1,128 @@
...
+      </dependency>
+
+      <!-- Import the JAX-RS API, we use provided scope as the API is included 
+         in JBoss AS 7 -->
+      <dependency>
+         <groupId>org.jboss.spec.javax.ws.rs</groupId>
+         <artifactId>jboss-jaxrs-api_1.1_spec</artifactId>
+         <scope>provided</scope>
+      </dependency>
+      <dependency>
```
pmuir: You should add some comments about this dep.

logging-tools-qs/pom.xml
```diff
@@ -0,0 +1,128 @@
...
+         <plugin>
+            <artifactId>maven-war-plugin</artifactId>
+            <version>2.1.1</version>
+         </plugin>
+
+         <!-- JBoss AS plugin to deploy war -->
+         <plugin>
+            <groupId>org.jboss.as.plugins</groupId>
+            <artifactId>jboss-as-maven-plugin</artifactId>
+            <version>7.1.0.Final</version>
```
jamezp: Might as well go do 7.1.1.Final here.

pmuir:
You need to use the IDE formatters at https://github.com/jboss/ide-config

logging-tools-qs/src/main/java/org/jboss/as/quickstarts/loggingToolsQS/DateService.java
```diff
@@ -0,0 +1,71 @@
...
+			days = (float)target.getTime() - now.getTime();  
+			days = days / (1000*60*60*24);	//turn milliseconds into days
+		} 
+		catch (ParseException ex) 
+		{
+			//create localised ParseException using method from bundle with details from ex
+			ParseException nex = DateExceptionsBundle.EXCEPTIONS.targetDateStringDidntParse(targetdate, ex.getErrorOffset());
+
+			//log a message using nex as the cause
+			DateLogger.LOGGER.logStringCouldntParseAsDate(targetdate, nex);
```
jamezp: I think it's good here to show the different ways to log messages and create exceptions. There should probably be a comment though indicating that this is verbose for the reasons of the example and it's generally no recommended to recreate exceptions or log exceptions that are being thrown. Kind of a disclaimer I guess :-)

logging-tools-qs/src/main/java/org/jboss/as/quickstarts/loggingToolsQS/DateService.java
```diff
@@ -0,0 +1,71 @@
...
+
+			//throw a WebApplicationException (400) with the localised message from nex
+			throw new WebApplicationException(Response.status(400)
+					.entity(nex.getMessage())
```
jamezp: This should probably be nex.getLocalizedMessage()

logging-tools-qs/pom.xml
pmuir: We need to do this across the board.

jamezp: I can do that if you'd like. I can do it on this branch, wait until this is merged or do it on the current master.

pmuir: Or just a new branch. It should merge fine.

jamezp:
Shouldn't this also be added as a module in the parent POM? I realize (realise) these modules don't really rely on the parent POM, but IDEA seems to think it's not a Java project without being a <module/> in the parent POM.

pmuir:
Yes it should.

jamezp:
We might want to hold off the Japanese translations unfortunately. There is a bug in the logging tooling and maybe somewhere else too. The logging tooling is a simple fix, but I'm not sure how fast we'll cut a release.

The other bug could be about anywhere. I need to dig into it.

sgilda:
Do any of the double-byte languages work with logging tools yet?

jamezp:
Probably not. We weren't loading the properties files with UTF-8 so the Unicode was bad in the generated source. I'm doing a fix for that now though.

sgilda:
I hope UTF-8 encoding is used in methods that require it too. :-)

jamezp:
Okay there is a work-around for the logging tooling. If we switch the Japanese to Unicode in the properties files, the logging works. Still a bug when displaying on the web page though.

jamezp:
Seems to work with these changes https://github.com/jamezp/quickstart/commit/110e788cb800939967eb7bb263d7381ac3215612. The output in Chrome doesn't look right in Japanese, but I think that's a browser issue. Using wget for Japanese looks fine.

dmison:
Opps, some conflicts prompted me into dropping this branch and republishing it.  

All the feedback has been taken into account.

https://github.com/dmison/quickstart/commit/3968a79e74872501dd372ee5ddf209995d2c1bd6 was the first commit after the feedback.

I've left the crashme method as is - added some comments to clarify that it is a very contrived demo.

sgilda:
Hi Darrin,
Looks good! 

Since you started this quickstart, the README file template has changed. I'm trying to make them follow a standard format and use action verbs for the sections. I've modified yours and sent it to you by email.

Minor comment: since most of the quickstarts run at http://localhost:8080/QUICKSTART_NAME, would it be possible to create an index page and redirect http://localhost:8080/jboss-as-logging-tools/ to  http://localhost:8080/jboss-as-logging-tools/rest/greetings/`name` by default? 

Also, is there any date verification for the daysuntil URL? When I enter: http://localhost:8080/jboss-as-logging-tools/rest/dates/daysuntil/25-25-2012 (which is an invalid date), it returns 661. :-)

Pete or James, do you need to review?

jamezp:
Looks good to me.

pmuir:
I'll review and merge today. Sande, do you want to send me the updated README for me to merge in?

pmuir:
Merged

dmison:
On 04/04/2012, at 6:44 AM, sgilda wrote:

> Minor comment: since most of the quickstarts run at http://localhost:8080/QUICKSTART_NAME, would it be possible to create an index page and redirect http://localhost:8080/jboss-as-logging-tools/ to  http://localhost:8080/jboss-as-logging-tools/rest/greetings/`name` by default? 

Because the deployed artefact is multiple service endpoints, I'm not sure about that.  Perhaps a index.html which contains a list of the endpoints ? 

> Also, is there any date verification for the daysuntil URL? When I enter: http://localhost:8080/jboss-as-logging-tools/rest/dates/daysuntil/25-25-2012 (which is an invalid date), it returns 661. :-)

Doh!  I tested some invalid dates, but only those that didn't fit the format string because I was interested in the ParseExceptions.  I'm surprised that it doesn't throw an invalid date exception.  I'll have a look.

pmuir:
I like the idea of an index which contains the list.

sgilda:
Me too. It gives a landing page and shows the quickstart has multiple 
endpoints/

PR #155, opened 2012-03-21
Updated to work on Openshift

paulrobinson:
This is now ready for review. 

There are a couple of untested tidy-ups as Openshift is currently down for maintenance:

OpenShift will be performing an update on Wednesday Mar 21st, some services may be unavailable.

### DIFF
--- a/wsat-simple/src/main/java/org/jboss/as/quickstarts/wsat/simple/servlet/WSATSimpleServletClient.java
+++ b/wsat-simple/src/main/java/org/jboss/as/quickstarts/wsat/simple/servlet/WSATSimpleServletClient.java
@@ -62,7 +62,7 @@ public class WSATSimpleServletClient extends HttpServlet {
         resp.setContentType("text/html");
         PrintWriter out = resp.getWriter();
 
-        out.write("<h1>Quickstart: Example demonstrates the use of *JMS 1.1* and *EJB 3.1 Message-Driven Bean* in JBoss AS 7.1.0.</h1>");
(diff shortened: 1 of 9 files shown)


wsat-simple/src/main/java/org/jboss/as/quickstarts/wsat/simple/servlet/WSATSimpleServletClient.java
```diff
@@ -0,0 +1,107 @@
...
+        BindingProvider bindingProvider = (BindingProvider) client;
+        List<Handler> handlers = new ArrayList<Handler>(1);
+        handlers.add(new JaxWSHeaderContextProcessor());
+        bindingProvider.getBinding().setHandlerChain(handlers);
+        //Lookup the DNS name of the server from the environment and set the endpoint address on the client.
+        String openshift = System.getenv("OPENSHIFT_APP_DNS");
+        if (openshift != null) {
+            bindingProvider.getRequestContext().put(BindingProvider.ENDPOINT_ADDRESS_PROPERTY, "http://"+openshift+"/wsat-simple/RestaurantServiceAT");
+        }
+
```
paulrobinson, line 61: I'm not happy with this part of the code. There should be a way of configuring the endpoint address through configuration. Maybe this is OK to leave in for the baseline?

pmuir, line 61: So basically what you are doing is relying on a system property that openshift always sets up?

Do we use this servlet now on localhost as well?

paulrobinson, line 61: Correct. 

You can use this Servlet locally too, but the readme only tells the user how to run the Arq test.

pmuir:
I updated the README to align with Mike's and merged. Great work!

PR #156, opened 2012-03-21
Template for creating a README.md file for new quickstarts

sgilda:
Adding this first. I don't think this one requires changes to the dist script do we? They may want to generate the HTML to see how it looks.


pmuir:
Yes, I needed to exclude the template directory. Merged.

PR #159, opened 2012-03-21
Fix instructions to deploy EARs

sgilda:
This addresses the deployment failures for ejb-in-ear and kitchensink-ear. Reworked the README files as long as I was in there to better match the template.

### DIFF
--- a/ejb-in-ear/README.md
+++ b/ejb-in-ear/README.md
@@ -9,16 +9,13 @@ This example demonstrates the deployment of an EAR artifact. The EAR contains: *
 
 The example is composed of three maven projects, each with a shared parent. The projects are as follows:
 
-1. ejb
-This project contains the EJB code and can be built independently to produce the JAR archive.
+1. ejb: This project contains the EJB code and can be built independently to produce the JAR archive.
 
-2. web
-This project contains the JSF pages and the managed bean.
+2. web: This project contains the JSF pages and the managed bean.
 
-3. ear
-This project builds the EAR artifact and pulls in the ejb and web artifacts.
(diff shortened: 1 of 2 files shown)


ejb-in-ear/README.md
```diff
@@ -33,68 +30,73 @@ displayed to the user.
...
 
-If you created the project using the Maven archetype wizard in your IDE
-(Eclipse, NetBeans or IntelliJ IDEA), then there is nothing to do. You should
-already have an IDE project.
+Import the project into an IDE
+---------------------
 
-Detailed instructions for using Eclipse with JBoss AS 7 are provided in the
-JBoss AS 7 <a href="https://docs.jboss.org/author/display/AS71/Getting+Started+Developing+Applications+Guide" title="Getting Started Developing Applications Guide">Getting Started Developing Applications Guide</a>.
+If you created the project using the Maven archetype wizard in your IDE (Eclipse, NetBeans or IntelliJ IDEA), there is nothing to do. You should already have an IDE project.
```
pmuir, line 90: This isn't right, you can't create this app from an archetype. I've reverted locally.

sgilda:
The deploy instructions still need to be changed though. The 2 separate commands don't work.
Sounds like I need to modify the template README also?

pmuir:
Yes, I've just reverted the IDE bit.

pmuir:
Merged. I fixed the indenting for code blocks to be consistent, and also added a bit of back tick action.

sgilda:
Pete, turns out that instruction was in the README before I edited it. Because it was moved, the diff showed it as being added.
It is not in the template version so there's nothing to fix. :-)

PR #162, opened 2012-03-23
Allow the java.naming.provider.url to be set externally

starksm64:


sgilda:
Hi Scott,
I just tried this and the 'mvn clean compile exec:java' fails with the following:

Mar 23, 2012 2:27:26 PM org.jboss.remoting3.remote.RemoteConnection handleException
ERROR: JBREM000200: Remote connection failed: javax.security.sasl.SaslException: Authentication failed: all available authentication mechanisms failed

Also, the system property you added to the pom.xml file was in the commented section of the system properties. Is that correct?

Should something be added to the README to explain what this property is and how it is used?

Thanks,
Sande

sgilda:
Sorry. Missed the add user step. Will add that. I think we need to move that instruction up before testing the quickstart. Similar to the template/README.md file.

pmuir:
Sande, once you get it working, please merge.

sgilda:
It does workl when you add the user. :-)

If you add information about the new java.naming.provider.url. system property to the README, could you replace the existing instructions to add the user with the following:

This quickstart uses secured management interfaces and requires that you create an application user to access the running application. Instructions to set up an application user can be found here: [Add an Application User](../README.html/#addapplicationuser)

pmuir:
Sande, jboss.naming.provider.url allows the JNDI directory used to lookup the JMS Destination to be configured. When running on localhost, it is normally on the default address, but if you are using the client on another host, you would need to specify the host to look it up on.

sgilda:
OK. I'll update the readme with this info. Thanks Pete.

starksm64:
Do you need to to update anything on this pull, or have you taken it over Sande?

sgilda:
Scott. This was fine. I made the changes to the README, then merged and closed. Thanks!

sgilda:
Merged and closed.

PR #164, opened 2012-03-26
Wicket

pmuir:
Replacement (part 1, ear only) for pull request #151. See my commit message for details.


sgilda:
I don't believe this quickstart can follow the standard build and deploy that is described in the README in the root directory. I get the same error I was getting in the ejb-in-ear and kitchensink-ear quickstarts. 

EARs require the one step "mvn clean package jboss-as:deploy". I should probably add that to the README in the root directory since a few of the quickstarts require it.

Also, the web-context is registered as: /jboss-as-wicket-ear-war
So the link in the README should be changed to: http://localhost:8080/jboss-as-wicket-ear-war

pmuir:
Ok, that's fixed. Merge when you think it's ready :-)

pmuir:
Merged

OndraZizka:
Seems you've done all the rest - or, anything else to be done?

BTW, there's a typo, see
https://github.com/pmuir/quickstart/commit/7cab9a19332666fc276652ddb9380139f067113b#commitcomment-1136964

PR #174, opened 2012-04-02
initial dom4j port of q40

baranowb:
note: q40 has been stripped of template def, but it still includes it in resources, is this on purpose?

### DIFF
--- a/xml-dom4j/pom.xml
+++ b/xml-dom4j/pom.xml
@@ -1,0 +1,129 @@
+<?xml version="1.0"?>
+<project
+	xmlns="http://maven.apache.org/POM/4.0.0"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+
+	<groupId>org.jboss.as.quickstarts</groupId>
+	<artifactId>jboss-as-xml-dom4j</artifactId>
+	<version>7.0.2-SNAPSHOT</version>
+	<packaging>war</packaging>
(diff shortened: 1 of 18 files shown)


baranowb:
This one also does not use template.

pmuir:
No, you could send a pull req to remove it.

xml-dom4j/pom.xml
```diff
@@ -0,0 +1,129 @@
+<?xml version="1.0"?>
```
pmuir, line 1: Could you indent by 4 spaces, just like Java files in all files?

xml-dom4j/pom.xml
```diff
@@ -0,0 +1,129 @@
...
+<project
+	xmlns="http://maven.apache.org/POM/4.0.0"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+
+	<groupId>org.jboss.as.quickstarts</groupId>
+	<artifactId>jboss-as-xml-dom4j</artifactId>
+	<version>7.0.2-SNAPSHOT</version>
```
pmuir, line 10: Version is now 7.1.1-SNAPSHOT :-D

pmuir:
I assume the only difference between this and xml-jaxp is the in DOM4JXMLParser?

baranowb:
Yeah, thought thats the point :) ? Like hibernate demos.

pmuir:
Yes. But you confirming that saves me checking it :-)

xml-dom4j/pom.xml
baranowb, line 10: hmm seems like other quickstarts need update also :)

pmuir:
I'll make the updates when I do the merge.

pmuir:
Merged

PR #181, opened 2012-04-03
Updated the version of jboss-javaee-6.0 to 3.0.0.Final

fjuma:


pmuir:
Looks good, Sande can you test and merge?

sgilda:
Pete: Is there a better way to test than to build , deploy, and run each one individually?

sgilda:
I tested a couple that had the change to jboss-javaee-6.0 to 3.0.0.Final and a couple of the jboss-jsf-api_2.1_spec changes and they work with no problem. Do I need to test all of them?

pmuir:
The only way to test is to do a full build and deploy on all.

pmuir:
Let me see if building from root will work.

pmuir:
Merged

PR #195, opened 2012-04-10
Separated producer from repository

mbogoevici:
- isolate list producer as a separate request-scoped bean
- made repository application-scoped


sgilda:
I tested this change and it works fine.

Do the archetypes need to be modified to remain in sync?
What about the other kitchensink quickstarts, like kitchensink-html5-mobile and kitchensink-jsp?

pmuir:
Yes, the archetypes need sync'ing, as do the other quickstarts. Marius, can you notify the other authors? And also issue the pull for the archetypes...

pmuir:
Merged.

PR #196, opened 2012-04-11
jax-rs-client

blainemincey:
Made slight modifications to the pom.xml and JaxRsClientTest class.  In the pom.xml, added the surefire plugin in order to read system properties from the pom.xml.  These values can now be modified in the pom in order to make this more testable when interacting with the helloworld-rs if it is deployed somewhere other than localhost like OpenShift for example where everyone's URL will be different.

Slight modification to the JaxRsClientTest class to read values from System properties that were added in the pom.xml.  A new method, beforeClass, was added in order for the System properties to be loaded prior to executing the test method.

### DIFF
--- a/jax-rs-client/pom.xml
+++ b/jax-rs-client/pom.xml
@@ -1,59 +1,70 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
-    <modelVersion>4.0.0</modelVersion>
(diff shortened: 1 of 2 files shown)


sgilda:
I tested and it works fine.

The only question I have is whether the version below should be '7.1.1-SNAPSHOT' (like the other quickstarts) instead of '1.0'
    `<groupId>org.jboss.as.quickstarts</groupId>`
    `<artifactId>jax-rs-client</artifactId>`
    `<version>1.0</version>`
    `<name>JBoss AS Quickstarts: JAX-RS Client</name>`
    `<description>jax-rs-client: Demonstrates the use an external JAX-RS RestEasy client</description>`

Pete, do you have any comments?

pmuir:
I merged this - fixing Sande's point along the way :-D

PR #208, opened 2012-04-13
Fixed issues multiple issues in pom.xml and README.md files

sgilda:
201: added missing Apache licenses 
203: fix wrong versions in pom files 
206: jta-crash-rec README fixes 
207: Replace anchor tags to be compatible with markdown


pmuir:
Merged

PR #210, opened 2012-04-13
Prep for OpenShift

blainemincey:
Made modifications to the README.md file to include instructions on how to deploy the helloworld-rs application to OpenShift and make the necessary modifications to the jax-rs-client pom.xml to execute against the deployed service.  I copied the helloworld-rs instructions for OpenShift and added the necessary information to edit the correct places in the pom.xml of jax-rs-client.

### DIFF
--- a/jax-rs-client/README.md
+++ b/jax-rs-client/README.md
@@ -78,3 +78,112 @@ MediaType: application/json
 
 ===============================================
 
+OpenShift
+
+To make this quickstart more interesting, deploy the RESTful service to OpenShift.  The following instructions will guide you as to the modifications that
+must be made to successfully execute the jax-rs-client against a service deployed to OpenShift.
+
+
+Build and Deploy the Quickstart - to OpenShift
+-------------------------
+
+### Create an OpenShift Express Account and Domain
+
+If you do not yet have an OpenShift Express account and domain, visit <https://openshift.redhat.com/app/login> to create the account and domain. The [OpenShift Express Getting Started Guide](http://docs.redhat.com/docs/en-US/OpenShift_Express/2.0/html/Getting_Started_Guide/index.html) will show you how to install the OpenShift Express command line interface.
+
+### Create the OpenShift Application
+
+Open a shell command prompt and change to a directory of your choice. Enter the following command:
+
+        rhc app create -a helloworldrs -t jbossas-7
+
+This command creates an OpenShift application called `helloworldrs` and will run the application inside a `jbossas-7` container. You should see some output similar to the following:
+
+    Creating application: helloworldrs
+    Now your new domain name is being propagated worldwide (this might take a minute)...
+    Warning: Permanently added 'helloworldrs-quickstart.rhcloud.com,107.22.36.32' (RSA) to the list of known hosts.
+    Confirming application 'helloworldrs' is available:  Success!
+    
+    helloworldrs published:  http://helloworldrs-quickstart.rhcloud.com/
+    git url:  ssh://b92047bdc05e46c980cc3501c3577c1e@helloworldrs-quickstart.rhcloud.com/~/git/helloworldrs.git/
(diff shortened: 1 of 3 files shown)


jax-rs-client/README.md
```diff
@@ -78,3 +78,112 @@ MediaType: application/json
...
+pom.xml.  Edit the values for xmlUrl and jsonUrl.  It should resemble the following:
+
+<property>
+        <name>xmlUrl</name>
+        <value>http://helloworldrs-quickstart.rhcloud.com/xml</value>
+</property>
+<property>
+        <name>jsonUrl</name>
+        <value>http://helloworldrs-quickstart.rhcloud.com/json</value>
+</property>
```
pmuir, line 153: My only comment here is that you could actually specify these as command line properties so that people can do `mvn test -DxmlUrl=... -DjsonUrl=..." - might be easier?

You would need to use a property replacement symbol in the pom.xml ${jsonUrl}, and set default values in the properties block.

jax-rs-client/pom.xml
```diff
@@ -1,59 +1,70 @@
...
 
-    <groupId>org.jboss.as.quickstarts</groupId>
-    <artifactId>jax-rs-client</artifactId>
-    <version>1.0</version>
-    <name>JBoss AS Quickstarts: JAX-RS Client</name>
-    <description>jax-rs-client: Demonstrates the use an external JAX-RS RestEasy client</description>
+	<groupId>org.jboss.as.quickstarts</groupId>
+	<artifactId>jax-rs-client</artifactId>
+	<version>1.0</version>
+	<name>JBoss AS Quickstarts:  jax-rs-client</name>
```
pmuir, line 8: Could you not randomly change whitespace? Just set your eclipse XML file editor to 4 spaces and do a Crtl-Shift-F to get rid of the problems :-)

pmuir:
Minor comments, otherwise looks great!

sgilda:
Hi Blaine,
I have a few minor comments. Did you rebase upstream? You're missing a couple of changes. 
For example, in the pom.xml file, the version was changed from '1.0' to '7.1.1.SNAPSHOT'.
Also, the README file changed. Pete removed the '-Parq-jbossas-remote' parameter from the 'maven clean test' command.
Other than that, it looks good. :-)
Thanks, 
Sande

blainemincey:
Hey folks.

Sande, I cloned from master.  Looks like I needed to grab what was upstream though.  Should I do that (how do I do that :) ) and if not,I can make the changes and submit a separate pull request.  Let me know how best to proceed.

Pete, I can make the changes you suggested for reading properties from the commandline using property substitution if that is your preference.  If I make that change, I will definitely have to re-do my video and not sure how much time I will have for that this week.  However, if that is what needs to happen, I can make it happen one way or another.

sgilda:
I'm not an expert at git, but I think you do this:

git fetch upstream
git rebase upstream/master

At this point you may see conflicts. If you do, fix them, then do a 'git 
add <filename>'  for each file you have to fix, then 'git rebase continue'.

After that, you git push to your repo (I think).

Pete, am I telling Blaine the correct process here?

pmuir:
No, it was really just a suggestion, I think it can work fine as it is today.

blainemincey:
Thanks for the instructions Sande.  Unfortunately, when I do a 'git fetch upstream' I get the following: fatal: 'upstream' does not appear to be a git repository.

Pete can you comment?  By the time I understand all the git commands, we will be on AS 10!

sgilda:
Sorry about that. I renamed my remote repository 'upstream'. 
When you do 'git remote -v' what is the name for your remote repository for https://github.com/jbossas/quickstart? You would use that name.

sgilda:
Blaine, I'm on IRC if you run into any problems.

pmuir:
I rebased and merged.

PR #218, opened 2012-04-20
Fix issue 214: build failure on jax-rs-client. 

sgilda:
Added helloworld-rs module dependency:

Pete, can you review this change for me. I tested it and it worked.


sgilda:
This does not seem to fix the problem after all. Closing this pull.

PR #225, opened 2012-05-01
Update_wsat-simple_to_use_default_context

paulrobinson:
Changed the local and Openshift parts of the quickstart to use the default url context, rather than the "wsat-simple" url context

### DIFF
--- a/wsat-simple/README.md
+++ b/wsat-simple/README.md
@@ -71,6 +71,8 @@ Run the Arquillian Tests
 
 This quickstart provides Arquillian tests. By default, these tests are configured to be skipped as Arquillian tests require the use of a container. 
 
+As we will be the application into the root context, we need to disable the welcome application. This can be done by editing './docs/examples/configs/standalone-xts.xml' and going to the 'urn:jboss:domain:web:1.1' subsytem. Here you will see the 'enable-welcome-root' attribute. Set this to false, to ensure the welcome application is not deployed.
+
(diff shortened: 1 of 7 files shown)


wsat-simple/README.md
```diff
@@ -71,6 +71,8 @@ Run the Arquillian Tests
 
 This quickstart provides Arquillian tests. By default, these tests are configured to be skipped as Arquillian tests require the use of a container. 
 
+As we will be the application into the root context, we need to disable the welcome application. This can be done by editing './docs/examples/configs/standalone-xts.xml' and going to the 'urn:jboss:domain:web:1.1' subsytem. Here you will see the 'enable-welcome-root' attribute. Set this to false, to ensure the welcome application is not deployed.
```
pmuir: Why do we need to do this?

pmuir:
Do we really want to do this for local as well?

wsat-simple/README.md
paulrobinson: This is because the welcome application uses the root context. You can't deploy an application into the root context if it is already in use. Shall I update the readme to make this clearer?

pmuir:
It's probably not a good to deploy this to the context root locally.

If we always need to deploy to the same context root, then we should change the openshift stuff IMO.

paulrobinson:
I think you're right, I'll try and revert the local part to not use the context root. I remember that I thought there was an issue with this, but from looking at the code now, I can't see what it is. I'll just give it a go and see if I can get it working.

wsat-simple/README.md
paulrobinson: This is now removed.

paulrobinson:
I've now reverted the local version back to using the wsat-simple context for local deployments.

pmuir:
Merged

PR #237, opened 2012-05-03
clarify that the domain portion of the OpenShift URL will be your own domain

sgilda:


sgilda:
Pete, I went ahead an moved this in. If you don't like it, I can always change it.

PR #242, opened 2012-05-07
Fixing broken urls at the main page

abstractj:
All the URLs to the projects at the main page are broken. This is gonna fix the github issue (https://github.com/github/markup/issues/84)


pmuir:
This will break all the links locally :-(

abstractj:
Are we running README.md locally?

pmuir:
Yes, every release :-)

On 7 May 2012 23:32, Bruno Oliveira
reply@reply.github.com
wrote:

abstractj:
Got it :)

PR #294, opened 2012-07-09
Inter app

pmuir:


sgilda:
Pete, this is a very cool example. I have a few minor comments for the README:

Change this: 
  This quickstart shows you how to easily communicate between two modular deployments to JBoss AS 7. 
To:
   JBoss Enterprise Application Platform 6 or JBoss AS 7. 

You say: In all, the project has three modules:
But you only list 2. :-)

I originally thought it would be nice to have 2 buttons: 1 to send the message and 1 to update the current screen, so you could update without sending any message, but it doesn't look like it would be easy to do. Besides, this works great as it is.

sgilda:
Did you try an undeploy? I'm having a few issues but it could be something I'm doing. I'll try again with a fresh server.

sgilda:
Looks good!

sgilda:
Merged.

PR #297, opened 2012-07-13
Fix issue 295: remove github URLs from the parent POM

sgilda:
Removed obsolete github URLs from POM. Added hbernate3 as issue is resolved.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,8 +24,8 @@
     <packaging>pom</packaging>
     <name>JBoss AS Quickstarts Parent</name>
     <description>JBoss AS Quickstarts Parent</description>
-    <url>http://github.com/jbossas/quickstart</url>
-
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -24,8 +24,7 @@
     <packaging>pom</packaging>
     <name>JBoss AS Quickstarts Parent</name>
     <description>JBoss AS Quickstarts Parent</description>
-    <url>http://github.com/jbossas/quickstart</url>
```
pmuir, line 27: Actually, why don't we provide http://jboss.org/jdf/quickstarts/get-started as the URL here?

pmuir:
Looks good.

sgilda:
Excellent idea to use the jdf URL. Modified and squashed.

pmuir:
Merged

PR #299, opened 2012-07-15
jboss-jdf_296 wsat-simple uses obsolete github URL for namespaces

paulrobinson:
I've changed all namespaces to "www.jboss.com/jboss-jdf/jboss-as-quickstart". Let me know if this was the wrong namespace to change to. I've stored the command to do the replace, so it's a quick fix to change again.


pmuir:
Paul, I think it needs to jboss.org, not jboss.com

paulrobinson:
Yeah, I agree. Will fix shortly...

paulrobinson:
@pmuir done.

sgilda:
I'm sorry to ask a very stupid question, but do the namespace links have to be valid?
For example, this link gets a Not Found: http://www.jboss.org/jboss-jdf/jboss-as-quickstart/wsat/simple/Restaurant

I assume it's just a unique namespace and doesn't need to be a valid URL, but since there was so much discussion about changing it, I want to be sure. :-)

pmuir:
Yeah, a namespace is just a namespace. This is why I prefer urns, as they are obviously not valid URLs.

sgilda:
Yes. It would be less confusing if it was an URN. :-)

This looks fine then. Do you want me to push it?

pmuir:
Please

sgilda:
Pushed.

PR #306, opened 2012-07-19
Issue 270

sgilda:
Fix issue #270:Shopping cart has wrong element name in commented XML.


helloworld/src/main/java/org/jboss/as/quickstarts/helloworld/HelloWorldServlet.java
```diff
@@ -52,6 +52,7 @@
 
    @Override
    protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws ServletException, IOException {
+      resp.setContentType("text/html");
```
pmuir, line 55: Sande, for this one, we should update the guide as well. Do you want to try doing that, so you get familiar with it?

pmuir:
I've merged in the fix for #270

pmuir:
Ah, i see we just have a bit of pull req confusion, I'll close this one then.

PR #312, opened 2012-07-25
Fix issue #269: Add jboss-javaee-6.0-with-tools dependency

sgilda:
To hibernate4. . POM should be similar to kitchensink.


pmuir:
Replaced with cd269e75fc6e2235b7db33f0e0d7328be1441601

PR #313, opened 2012-07-25
Fix issue #277: remove hibernate-validator and junit from kitchensink-ear parent POM

sgilda:
Pete, I'm not sure this is the correct solution. 
PLEASE VERIFY THIS CHANGE before merging!!!


pmuir:
Merged

PR #314, opened 2012-07-25
Partial fix for issue #245: apply dual branding for hibernate3

sgilda:
Pete, this seems to look like the kitchensink and hibernate4 quickstarts, but please check this carefully before you merge to make sure I did it correctly! ;-)


sgilda:
Withdrawing this. I'm seeing an exception so I must have missed something.

sgilda:
Fixed the index.xhtml file. This seems to be working now. 
Pete: Please do a thorough check of this though!

sgilda:
Just discovered the labels aren't right. They're centered and wrap. Hold off on this for a bit.

sgilda:
It looks like the hibernate4 (where I stole the code) has the same issue with the labels. I need to investigate both.

sgilda:
Pete: 
In the index.xhtml for hibernate4, there's this code:
    `<h:panelGrid columns="3" columnClasses="titleCell">
            <h:outputLabel for="id" value="Member Id (Enter unique number):"/>
            <h:inputText id="id" value="#{newMember.id}" />
            <h:message for="id" errorClass="invalid"/>`
The labels seem to be centered and the column is too narrow, so text wraps.
I don't see "titleCell" in the screen.css. 
How do I: 
- left align the labels
- increase the width of the label column (although maybe I won't have to if the alignment is fixed

pmuir:
Sande, let me know when this is ready to merge

sgilda:
I think this is ready now. I'll apply the same label display fix to hibernate4 if you're ok with it.

pmuir:
Merged

PR #315, opened 2012-07-26
Partial fix for issue #245: Dual branding for kitchensink-jsp

sgilda:
I haven't done much JSP coding, so this needs to be checked. The deployed quickstart looks similar to the others.


sgilda:
This is missing the header background and possibly a few other images, but it's better than it was.

pmuir:
Ok, I finished it off :-)

PR #321, opened 2012-08-01
copy'n'paste accident

wfink:
depend on the bug AS7-5218 the stop() was not tested.
The stop() method call initialize instead of stop() for the scheduler
Thanks Paul for the hint during his test.


sgilda:
Tested this and it works fine.
Pete, do you have any questions or comments? If not, I'll merge it.

PR #326, opened 2012-08-06
Unify module names style.

OndraZizka:
Changes few modules' <name> to be like "Kitchensink JSF" instead of "kitchensink-jsf".


pmuir:
Thanks, this is excellent! Merged.

PR #346, opened 2012-10-17
Rename module folders to match their artifactId

fbricon:
As discussed on the jdf ML, the module folder names need to be renamed to match their artifact id, so that Eclipse WTP based tooling (JBoss Tools/ JBDS) can deploy the webapps using the proper context root.


fbricon:
There was a problem removing some modules from the git index. Let me try again

PR #347, opened 2012-10-17
Rename module folders to match their artifactId

fbricon:
As discussed on the jdf ML, the module folder names need to be renamed to match their artifact id, so that Eclipse WTP based tooling (JBoss Tools/ JBDS) can deploy the webapps using the proper context root.

This time I didn't lose any files :-)


sgilda:
@fbricon , were these changes made to the Aerogear quickstarts?

fbricon:
there are 2 html5 projects renamed here
https://github.com/fbricon/jboss-as-quickstart/tree/e128eb0a730d560611f0b591b639d9e8c1573b05
are we missing other projects?

sgilda:
@fbricon : I'm not exactly clear what you mean by "2 html5 projects", but I do have another question:

Is there a reason the `bean-validation`, `jts-crash-recovery`, and `kitchensink-cordova` were not renamed?

I can understand not renaming `forge-from-scratch` or `h2console`, although they look a little out of place right now.

fbricon:
On Tue, Oct 23, 2012 at 8:39 PM, sgilda notifications@github.com wrote:

> @fbricon https://github.com/fbricon : I'm not exactly clear what you
> mean by "2 html5 projects", but I do have another question:
> 
> html5 projects == aerogear projects
> 
> Is there a reason the bean-validation, jts-crash-recovery, and
> kitchensink-cordova were not renamed?
> 
> requirements (in the script) to rename the projects :
> - must have a pom.xml
> - folder != artifactId

So those projects don't meet all the requirements. The rename thing is
required for maven projects we import in eclipse. I didn't try to enforce
the rename when it was not absolutely necessary. But if you want them
renamed, for the sake of consistency, I can do that

> I can understand not renaming forge-from-scratch or h2console, although
> they look a little out of place right now.

sgilda:
The bean-validation quickstart should be using the jboss-as prefix in the artifact id. Good catch on your part! 
The jts-distributed-crash-recovery quickstart has no code. It builds on the deployed `jts` quickstart.
I'm not clear on the kitchensink-cordova.
I guess we should leave them as they are.
Thanks for the clarification!
Sande

sgilda:
There was some discussion on IRC about this and the general consensus was to kill this change.

From Jay Balunas:
For JDF in general there were some concerns around the impact on the site, merge process, and how sweeping the changes were.  I'll let Sande, Rafael and/or Pete discuss those more.

For AeroGear quickstart changes (and I think JDG) the workflow we need is to have these changes in the upstream repos (aerogear/as-quickstarts) and then merged to jdf through a tag there.  This is needed so we keep "upstream" the repository of record.  So to get this in AeroGear we'll need a jira and a PR there.  They we can get the merge down to JDF.

However before doing that I think we should discuss the impact of these changes.  We'll need to also update the README's, docs, wiki pages, etc...  

From me:
I think there is a larger impact than we realize if we change the quickstart folder names. The names would no longer match the names on the jdf-site or on jboss.org. Every place we refer to the name, for example, kitchensink, we would need to change it to jboss-as-kitchensink. This impacts README files, tutorials, community docs, the jdf-site, and I'm not sure what else. We would also need to change the metadata in the README files so the tables build correctly. Plus, the quickstart name is referred to throughout each README.

There are also some quickstarts with no pom.xml file like forge-from-scratch, jts-crash-recovery, h2console, and others that look a little strange without the 'jboss-as-' prefix.

Closing this pull for now.

PR #352, opened 2012-11-02
Ensure that the user knows to set the node-identifier for JTS

tomjenkinson:
There is an error in the current version of the quickstarts where the user is not notified to change the node-identifier. Although it works for this quickstart without setting this, in most cases the user should set it and so I recommend we change the guidance here

### DIFF
--- a/jts/README.md
+++ b/jts/README.md
@@ -98,10 +98,11 @@ You can modify the server configuration using the JBoss CLI tool or by manually 
     
         For Linux: bin/jboss-cli.sh --connect
         For Windows: bin\jboss-cli.bat --connect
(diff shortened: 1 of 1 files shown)


sgilda:
Hi Tom,

Looks great! One minor thing. I believe you need to add instructions to these sections:to remove the unique identifier when they are done testing:
- Remove the JTS Server Configuration using the JBoss CLI Tool
      The command is `/subsystem=transactions/:undefine-attribute(name=node-identifier)`
- Remove the JTS Server Configuration Manually

Other than that, it looks good to me!

sgilda:
Thanks Tom. I merged the fix.

PR #354, opened 2012-11-11
[WICKET-4785] [AS7-4554] AS 7 + Wicket: ClassNotFoundException: org.jbos...

OndraZizka:
...s.msc.service.ServiceName on deserialization

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ bin
 *.log
 .clover
 *.swp
+!/wicket-war/src/main/resources/META-INF/MANIFEST.MF
--- a/wicket-ear/pom.xml
+++ b/wicket-ear/pom.xml
@@ -55,7 +55,7 @@
         <!-- <version.org.jboss.spec.jboss.javaee.6.0>3.0.0.Final-redhat-1</version.org.jboss.spec.jboss.javaee.6.0> -->
 
         <!-- Other dependency versions -->
-        <version.org.apache.wicket>1.5.5</version.org.apache.wicket>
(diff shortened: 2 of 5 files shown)


sgilda:
Hi Ondra,
I assume this is a fix for some bug? I have a couple of questions:

How do I test this to make sure the fix solves the problem? In other words, how would I get it to break in the current version?

Would you mind squashing the commits to make it easier to review the changes?

Thanks,
Sande

OndraZizka:
Hi Sande, yes it's a bug fix. To try, start AS and deploy the quickstart; then fiddle with the app for a while, adding and deleting some items. Then press the back button. In this scenario, you should get CNFEx coming from deserialization, see the jiras.

OndraZizka:
I can squash, but:
1) is for EAR QS
2) is for WAR QS (different one)
3) is upgrading Wicket framework version - unrelated
4) is unrelated change in pom, clearing some garbage which doesn't belong there

So I suggest to squash 1 + 2  and  3 + 4, wdyt?

sgilda:
Sorry. I didn't realize it affected multiple quickstarts! Sorry about that. Leave it as is.

Also, thanks for the explanation!

sgilda:
I tested the wicket-ear and wicket-war quickstarts with this change and they work fine.

Pete, this looks good to me. Do you need to review these changes? There are no code changes, only dependencies.

pmuir:
Looks good.

sgilda:
Merged.

PR #356, opened 2012-11-12
New Spring based WebsphereMQ quickstart

doug-grove:
hope for the best


sgilda:
There is a discussion on the jdf-list about whether this quickstart belongs here due do the need to purchase 3rd party software, complexity of configuration, and support issues.

sgilda:
Based on the feedback, I don't think this quickstart belongs here. It would probably be better as a GSS quickstart. Closing the pull.

PR #359, opened 2012-11-21
KitchenSink quickstart with DeltaSpike @Transactional instead of Contain...

btison:
...er Managed Persistence


sgilda:
@btison : One quick comment. Would you mind changing the name and folder to kitchensink-deltaspike? When I see "ds", I think datasource. :-)

sgilda:
This is great!

@pmuir: I'm good with this one. Do you need to review?

pmuir:
Yes I do. I've started to switch out dependency versions for managed versions at https://github.com/pmuir/quickstart/tree/kitchensink-ds but I've hit an issue with ShrinkWrap Resolver, as it's not clear whether we are ready to recommend this to customers, and whether it is a tested part of our product offering. I've emailed @kpiwko and asked him for comment.

pmuir:
Also, if we are positioning this as "kitchensink for DeltaSpike", we should make sure we use all relevant features such as exception handling.

pmuir:
I've update to managed versions and merged

PR #362, opened 2012-11-26
JDF - 147 - Upgrade i8ln quickstarts to use DeltaSpike i8ln

rafabene:
updated kitchensink-ml\* to use Deltaspike i18n


sgilda:
@rafabene , @pmuir : Do you want to replace the plain vanilla i18n quickstarts with ones that use DeltaSpike? That's fine if you do, but it offering separate quickstarts provides a side-by-side comparison. I also wonder if everyone will want to add DeltaSpike to their applications.

If you do want to replace them, I think the README files need to be updated to indicate that the kitchensink-ml and kitchensink-ml-ear use DeltaSpike and what to look for in the applications.

sgilda:
@pmuir: Is DeltaSpike supported in EAP 6.1? One concern is that these quickstarts ship with EAP 6.0. If we modify the existing quickstarts to add DeltaSpike rather than create separate quickstarts, does this affect the merge to EAP 6.1?

pmuir:
@sgilda yeah, that would move the quickstarts to WFK... Do you think that is a problem?

sgilda:
Would we need to remove the current ones from EAP 6.1 then? I assume we would leave them for EAP 6.0.1 though?

rafabene:
This pull request is deprecated since we'll use jboss logging.

PR #372, opened 2012-12-10
JDF-144 Using DeltaSpike for properties configuration

rafabene:


sgilda:
I tested this and it works fine. 
Pete, are you good with this?

sgilda:
After giving hthis one more thought, I prefer we leave the original quickstart as is, and copy and paste this as a new quickstart. 

@rafabene , @pmuir : Your thoughts?

rafabene:
Deprecated. We will add a new the quickstart instead of updating the existing one

PR #376, opened 2012-12-12
Issue 373 - Add instructions for root mvn install command

sgilda:


pmuir:
@sgilda might be nice to list all the possible profiles, and what they are for/require?

sgilda:
Good idea! I may need some help with a few of them though. I'll ping you if I need you. :-)

sgilda:
Added information about the Maven profiles defined in the root POM file.

pmuir:
Merged

PR #380, opened 2012-12-14
JDF-158 Add custom authorization example using @SecurityBindingType from DeltaSpike 

rafabene:


deltaspike-security/src/main/java/org/jboss/as/quickstarts/deltaspike/security/ErrorController.java
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * @author <a href="mailto:benevides@redhat.com">Rafael Benevides</a>
+ * 
+ */
+//The @Model stereotype is a convenience mechanism to make this a request-scoped bean that has an
+//EL name
+//Read more about the @Model stereotype in this FAQ:
+//http://sfwk.org/Documentation/WhatIsThePurposeOfTheModelAnnotation
+@Model
```
pmuir, line 36: No need to make this request scoped...

pmuir:
I made some changes and merged

PR #381, opened 2012-12-18
Fix JDF-95

sgilda:
Clear contents of toc.html file before each run to prevent growing quickstart table size
Move the maxdepth argument before type in the subdirectory find command


pmuir:
Merged

PR #405, opened 2013-01-22
Fix for JDF-183 - Arquillian tests hang, updated BOM version

sgilda:


sgilda:
For some reason, `tasks` still hangs. I need to check into that separately, but I wanted to get this checked in.

rafabene:
PR MErged

PR #408, opened 2013-01-22
Fix for issue 404 - meet new minimum password requirements

sgilda:


sgilda:
@rafabene : Good catches. I believe they are now all fixed. :-)

PR #417, opened 2013-01-29
adding cdi-alternative files round 2

nzhu972:
I created a new branch to submit the cdi-alternative quickstart example because there are problems on the other branch which is preventing me from pushing the changes.

The project is structured in a single WAR file and followed the standards outlined in the jdf contribution page.

### DIFF
--- a/cdi-alternative/README.md
+++ b/cdi-alternative/README.md
@@ -1,0 +1,99 @@
+cdi-alternative: Demostrates CDI Alternatives
+======================================================
+Author: Nevin Zhu
+Level: Intermediate
+Technologies: CDI, Servlet, JSP
+Summary: Demonstrates the use of CDI Alternatives where the bean is selected during deployment 
+
+What is it?
+-----------
+
+When more than one version of a bean is implemented for different purposes, the ability to switch between the versions during the development phase by injecting one qualifier or another is shown in this demo.
+
+Instead of having to change the source code of the application, one can make the choice at deployment time by using alternatives.
+
+Alternatives are commonly used for purposes like the following:
+
+1)  To handle client-specific business logic that is determined at runtime
+2)  To specify beans that are valid for a particular deployment scenario (for example, when country-specific sales tax laws require country-specific sales tax business logic)
+3)  To create dummy (mock) versions of beans to be used for testing
+
+To make a bean available for lookup, injection, or EL resolution using this mechanism, give it a javax.enterprise.inject.Alternative annotation and then use the alternative element to specify it in the beans.xml file.
+
+
+System requirements
+-------------------
+
+All you need to build this project is Java 6.0 (Java SDK 1.6) or better, Maven 3.0 or better.
+
+The application this project produces is designed to be run on JBoss Enterprise Application Platform 6 or JBoss AS 7. 
+
+ 
+Configure Maven
+---------------
+
+If you have not yet done so, you must [Configure Maven](../README.md#mavenconfiguration) before testing the quickstarts.
+
+
+Start JBoss Enterprise Application Platform 6 or JBoss AS 7
+-------------------------
+
+1. Open a command line and navigate to the root of the JBoss server directory.
+2. The following shows the command line to start the server with the web profile:
+
+        For Linux:   JBOSS_HOME/bin/standalone.sh
+        For Windows: JBOSS_HOME\bin\standalone.bat
+
+
+Build and Deploy the Quickstart
+-------------------------
+
+_NOTE: The following build command assumes you have configured your Maven user settings. If you have not, you must include Maven setting arguments on the command line. See [Build and Deploy the Quickstarts](../README.md#buildanddeploy) for complete instructions and additional options._
+
+1. Make sure you have started the JBoss Server as described above.
+2. Open a command line and navigate to the root directory of this quickstart.
+3. Type this command to build and deploy the archive:
+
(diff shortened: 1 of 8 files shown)


cdi-alternative/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+When more than one version of a bean is implemented for different purposes, the ability to switch between the versions during the development phase by injecting one qualifier or another is shown in this demo.
+
+Instead of having to change the source code of the application, one can make the choice at deployment time by using alternatives.
+
+Alternatives are commonly used for purposes like the following:
+
+1)  To handle client-specific business logic that is determined at runtime
+2)  To specify beans that are valid for a particular deployment scenario (for example, when country-specific sales tax laws require country-specific sales tax business logic)
+3)  To create dummy (mock) versions of beans to be used for testing
```
sgilda: These don't have line breaks for the numbered list when rendered in HTML. Try this with only 1 space between the period after the number and the start of the sentence. I would also get rid of the parentheses. 
1. To handle client-specific business logic that is determined at runtime.
2. To specify beans that are valid for a particular deployment scenario, for example, when country-specific sales tax laws require country-specific sales tax business logic.
3. To create dummy  or mock versions of beans to be used for testing.

cdi-alternative/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+        mvn clean package jboss-as:deploy
+        
+4. This will deploy `target/ear/cdi-alternative.ear` to the running instance of the server.
+
+
+Access the application
+---------------------
+
+The application will be running at the following URL <http://localhost:8080/cdi-alternative/Demo>.
```
sgilda: Most of the other quickstarts are accessed using the base URL, for example, http://localhost:8080/cdi-alternative/. That gives a 404 NOT FOUND. Is it possible to redirect it to the Demo page?

cdi-alternative/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+A message will display for the bean being injected during run time. 
+
+The alternative version of the beans can be updated in /WEB-INF/beans.xml by removing the <alternative> tag 
+
+or 
+
+change the class name to a different alternative.
```
sgilda: This is hard to read. Maybe something like this:

You can specify alternative versions of the bean in the WEB-INF/beans.xml file by doing one of the following:
1. You can remove the `<alternatives>` tag.
2. You can change the class name.

cdi-alternative/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+The alternative version of the beans can be updated in /WEB-INF/beans.xml by removing the <alternative> tag 
+
+or 
+
+change the class name to a different alternative. 
+
+In this quickstart, in order to switch back to the default implementation, 
+comment the <alternative> block. 
+
```
sgilda, line 75: I think you have to put tick marks around the `<alternatives>` tag. It's plural, plus, it doesn't appear in the text. :-) 
You also need to redeploy the application. 

To switch to the default implementation, comment the `<alternatives>` block in the WEB-INF/beans.xml file and redeploy the quickstart.

sgilda:
Tested this and it works great! Other than the few minor comments above, I'm good with this one.

@pmuir or @LightGuard : Do you want to review the code?

cdi-alternative/src/main/java/org/jboss/as/quickstarts/cdi/alternative/Demo.java
```diff
@@ -0,0 +1,73 @@
...
+     * If beans.xml has alternative defined, then that class will be injected.  
+     * Otherwise, the default is injected.
+	 */
+    @Inject
+    private Tax tax;       
+    /**
+     * default constructor
+     */
+    public Demo() {
+        super();
```
LightGuard: Not that the ctor hurts anything, but it certainly isn't needed.

cdi-alternative/README.md
```diff
@@ -0,0 +1,99 @@
...
+Instead of having to change the source code of the application, one can make the choice at deployment time by using alternatives.
+
+Alternatives are commonly used for purposes like the following:
+
+1)  To handle client-specific business logic that is determined at runtime
+2)  To specify beans that are valid for a particular deployment scenario (for example, when country-specific sales tax laws require country-specific sales tax business logic)
+3)  To create dummy (mock) versions of beans to be used for testing
+
+To make a bean available for lookup, injection, or EL resolution using this mechanism, give it a javax.enterprise.inject.Alternative annotation and then use the alternative element to specify it in the beans.xml file.
```
LightGuard: This isn't really accurate. Any Java class which has a no-args constructor and is in an archive with a beans.xml is available for lookup and injection. For EL resolution it must contain `@Named`

LightGuard:
I added a couple of comments.

nzhu972:
I updated README and the servlet based on both of your feedback. The page can also be accessed via http://localhost:8080/cdi-alternative as suggested instead of /Demo.

sgilda:
Thanks!. I'll try to look at it tomorrow.

sgilda:
I just realized the README is missing the `Target Product` metadata tag, however. I can add it later when I fix some other things. or before I merge.

Looks good to me.

@LightGuard : Let me know if you're OK with this and I'll merge it.

LightGuard:
Looks fine to me

sgilda:
Merged upstream.

PR #419, opened 2013-01-31
Update QS to use -with-resteasy BOM

rafabene:
Update jax-rs-client to use the new -with-resteasy BOM. This makes necessary to use 1.0.4.CR8 BOM that was updated on all QSs


sgilda:
Tested and merged.

PR #425, opened 2013-02-05
AS7-6453 and README.md clean-up

kwart:


ejb-security-interceptors/src/main/resources/META-INF/jboss-ejb3.xml
```diff
@@ -15,7 +15,7 @@
   ~ See the License for the specific language governing permissions and
   ~ limitations under the License.
   -->
-<jboss xmlns="http://www.jboss.com/xml/ns/javaee"
+<ejb-jar xmlns="http://www.jboss.com/xml/ns/javaee"
```
jaikiran: This isn't right. Why was this change required?

sgilda:
Bug 911483 was entered for this. I will hold off doing anything until Jaikiran's question is answered.

sgilda:
Please remove the metadata tag edits and the carriage returns that were added to all the text. :-)

ejb-security-interceptors/src/main/resources/META-INF/jboss-ejb3.xml
kwart: It's weird. I had issues with creating SwitchIdentityTestCase in AS7 testsuite and I was convinced it's related to this. But I'm not able to reproduce it now. So I'll change it back. Sorry.

kwart:
I hope the `README.md` is fixed now.

sgilda:
@kwart : I noticed you changed the formatting for all the top level sections in the file. It's not really a big deal except that we're trying to make sure all the quickstarts use the same formatting, using the template/README.md file as a guide, at least for common top level sections. This will allow us to more easily parse and scan to automate validation of things we expect to see in the file. If we change them in this file, we would need to update them in all the other README files to keep them consistent. I actually prefer the `#` symbols, but I'm  too busy (and lazy) to change all the files at this point.

Would you mind changing the top level section headers, including the first line, back to the original format? I'm not concerned about the subsections.

kwart:
should be returned now.

sgilda:
The README changes are good. 

@pmuir: Do you want to review the code or should I ask Darran or Jaikiran to review it?

@kwart: If I get the OK, I assume these changes need to go into the EAP 6.1 quickstarts too?

kwart:
Yes, it should be merged to the EAP 6.1 too .

sgilda:
@kwart, @darranl: While testing this for EAP 6.1, I noticed that under "Modify the Server Configuration using the JBoss CLI Tool", it tells you to start the server with `JBOSS_HOME_SERVER_1/bin/standalone.sh -c standalone-full.xml`. However, under "Start JBoss Enterprise Application Platform 6.1 or JBoss AS 7.2", it says to start the server using `JBOSS_HOME/bin/standalone.sh` without the argument. Does this quickstart require the full profile? I assume it does?

kwart:
@sgilda: No. AFAIK the default profile should be fine too.

sgilda:
@kwart: When you updated the README file, you added the full profile argument to the command to start the server. Later, when you start the server, you don't pass the profile that contains the server configurations. Don't they have to be in sync?

darranl:
Yes there is nothing HA specific in this example, I think standalone-full.xml crept in copying the CLI commands from a previous quick start.

sgilda:
Merged.

PR #427, opened 2013-02-06
Cluster-wide Singleton Bean Quickstart

rhusar:
PLEASE DO NOT MERGE

This quickstar demonstrates how to implement a clusterwide singleton bean on top of SingletonService in AS 7.1.3.Final. One can find some raw information in the readme.

I just want people to be able to find this quickstart and give feedback.

### DIFF
--- a/cluster-singletonbean/README.md
+++ b/cluster-singletonbean/README.md
@@ -1,0 +1,76 @@
+Configuration
+=============
+
+Unzip AS 7.1.3
+
+Create user
+./bin/add-user.sh
+
+application user
+ejb
+ejbejb
+
+
+Add to security-realms:
+
+            <security-realm name="ejb-security-realm">
+                <server-identities>
+                    <secret value="ZWpiZWpi"/>
+                </server-identities>
+            </security-realm>
+
+
+To remoting subsystem:
+
+            <outbound-connections>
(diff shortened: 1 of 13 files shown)


sgilda:
@rhusar : I'm not sure if you're looking for a review or not, but I think some things have changed since the pull was issued.
- The password in the README no longer meets the password rules. It has to be 8 characters, contain 1 digit and 1 special character. We also define a `quickstartUser/quickstartPwd1!` to use with the other quickstarts if you want to piggyback on that. However, if it uses other than the `ApplicationRealm`, that won't work.
- From the server configuration, it looks like the user `ejb` should be in the `ejb-security-realm`, not the `ApplicationRealm`.

When I tried the install, it fails with "Remote connection failed: javax.security.sasl.SaslException: Authentication failed: all available authentication mechanisms failed".

I tried changing the ejb user to the `ejb-security-realm`, but had no better luck. I might have misconfigured the server. I'll look into it a little more.

sgilda:
@rhusar : Just checking on the status of your quickstart. Have you had a chance to look at my comments?

sgilda:
Spoke with Radoslav on 8/2 and he mentioned he may close this pull as it not really production ready at this point. He add comments here in a few days.

rhusar:
Closing -- we need a proper way to support cluster-wide singletons. We will have this in WildFly 8/9 timeframe.

PR #428, opened 2013-02-06
Fix root README and helloworld-jms instructions

sgilda:


LightGuard:
Did this get changed to only support queues?

sgilda:
It turns out it never supported topics. :-)

PR #429, opened 2013-02-06
Fix arquillian.xml to match the other quickstart arquillian.xml files

sgilda:


LightGuard:
I don't see any problems

sgilda:
Merged.

PR #433, opened 2013-02-11
Fix pom file versions and comments

sgilda:
Please review carefully! 
I will create bugs for upstream quickstarts affected by these changes.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -38,11 +38,32 @@
         <!-- A base list of dependency and plugin version used in the various quick starts. -->
 
         <!-- JBoss dependency versions -->
+
         <version.org.jboss.as>7.3.Final</version.org.jboss.as>
-        <version.org.jboss.as.plugins.maven.plugin>7.3.Final</version.org.jboss.as.plugins.maven.plugin>
+        <!-- Alternatively, comment out the above line, and un-comment the 
+            line below to use version 7.1.1.Final-redhat-1 which is a release certified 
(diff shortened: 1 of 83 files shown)


pom.xml
```diff
@@ -240,6 +268,7 @@
             <modules>
                 <module>forge-from-scratch</module>
                 <module>h2-console</module>
+                <module>kitchensink-cordoba</module>
```
rafabene: it is cordova instead of cordoba

pom.xml
```diff
@@ -193,6 +220,7 @@
             </activation>
             <modules>
                 <module>cluster-ha-singleton</module>
+                <module>helloworld-jdg</module>
```
rafabene: I think helloworld-jdg could be on default profile. I don't think it is complex-dependencies

pom.xml
```diff
@@ -240,6 +268,7 @@
             <modules>
                 <module>forge-from-scratch</module>
                 <module>h2-console</module>
+                <module>kitchensink-cordoba</module>
                 <module>jts-distributed-crash-rec</module>
```
rafabene: It' is cordova instead of cordoba.

rafabene:
Finished review. Had come comments.

sgilda:
@rafabene : Great catch on both of those! Turns out the helloworld-jdg was failing for the same reasons carmart and carmart-tx were failing. I also removed <module>cmt</module> from the default profile. It doesn't belong there. This is ready for review again. :-)

rafabene:
It seems fine!

PR #434, opened 2013-02-12
Standardize property names, fix arquillian test configuration

sgilda:


LightGuard:
I've looked through about a quarter of this and it looks fine to me.

sgilda:
Thanks @LightGuard !

PR #438, opened 2013-02-13
Apply fixes made to EAP for bugs 910704, 910732, 910882

sgilda:


LightGuard:
I see nothing wrong here, as long as the final name is lining up with the name of the quickstart (so as to not cause confusion)

sgilda:
Thanks @LightGuard

PR #452, opened 2013-02-20
Fix for JDF-212: Modify kitchensink-html5-mobile pom file to use with-tools

sgilda:
...ols

### DIFF
--- a/kitchensink-html5-mobile/pom.xml
+++ b/kitchensink-html5-mobile/pom.xml
@@ -53,12 +53,7 @@
             maven repository. -->
         <!-- <version.jboss.spec.javaee.6.0>3.0.2.Final-redhat-1</version.jboss.spec.javaee.6.0> -->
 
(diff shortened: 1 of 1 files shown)


sgilda:
@rafabene : Actually, I wonder if I should add the hibernate dependency:

```
        <dependency>
            <groupId>org.jboss.bom</groupId>
            <artifactId>jboss-javaee-6.0-with-hibernate</artifactId>
            <version>${version.jboss.bom}</version>
            <type>pom</type>
            <scope>import</scope>
        </dependency>
```

 I see it in the Aerogear repository. What do you think?

rafabene:
I think that the upstream should remove -with-hibernate since it doesn't need it IMHO

rafabene:
Much better. I didn't see that hibernate dependencies were using explicit versions. In this case the last commit is more appropriate.

Ready to merge!

PR #455, opened 2013-02-25
Fix NoSuchEJBException in deltaspike-beanmanagerprovider

rsmeral:


rafabene:
Perfect! ;)

deltaspike-beanmanagerprovider/src/main/java/org/jboss/as/quickstarts/deltaspike/beanmanagerprovider/controller/ContactController.java
```diff
@@ -67,7 +67,6 @@
 
     private Contact contact;
     
-    @Inject
```
rafabene, line 70: After I merged I saw why I had to put the inject. When "onExceptionState" the list with allContacts gets clear because the ContactController was also destroyed when the conversation ended.

rafabene, line 70: Back @Inject annotation on http://github.com/jboss-jdf/jboss-as-quickstart/commit/d899047f3

PR #458, opened 2013-02-26
Removed Target Product from quickstarts that uses ShrinkWrap Maven Resolver

rafabene:
https://bugzilla.redhat.com/show_bug.cgi?id=915786


sgilda:
Merged.

PR #461, opened 2013-02-27
Fix bug 910704: Remove versions from archives

sgilda:
This duplicates the fix made to the EAP quickstarts.


LightGuard:
I don't see any issues with this

PR #465, opened 2013-03-05
Add element IDs to deltaspike-helloworld-jms

rsmeral:


sgilda:
This looks good to me. If no one objects, I will merge.

sgilda:
Based on the IRC conversation between Rafael and Ron, it looks like this is good to go. Merging.

PR #488, opened 2013-04-11
Fix the AngularJS Quickstart

wmeddie:
The AngularJS Kitchen Sink quickstart had the following problems:
1. Submitting an empty form caused a 415 error.
   - This is due to `newMember` never being initialized and AngularJS deleting the Content-Type header when posting an undefined object. (Safari defaults to "application/xml" and JAX-RS is only set to accept "application/json".)
2. Validation errors do not display.
   - The error callback of `Members.save` was defined as `function(data, status)` but is actually only called with a single parameter that has `data` and `status` fields.
3. The unknown errors list does not display.
   - On a 415 error the `Members.save` error handler set `$scope.errorMessages` but there is no corresponding slot in the view.
- I fixed 1 by adding a call to `$scope.reset()` after the call to `$scope.refresh()` thereby initializing the `newMembers` object on page load.
- I fixed 2 by changing the function to use the correct parameter.
- I fixed 3 by adding a `ul` that displays the `errorMessages` array if present.

### DIFF
--- a/kitchensink-angularjs/src/main/webapp/js/controllers.js
+++ b/kitchensink-angularjs/src/main/webapp/js/controllers.js
@@ -33,6 +33,8 @@ function MembersCtrl($scope, $http, Members) {
     $scope.register = function() {
         $scope.successMessages = '';
         $scope.errorMessages = '';
+        $scope.errors = {};
+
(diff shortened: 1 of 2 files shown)


pmuir:
Thanks, merged :-)

PR #489, opened 2013-04-11
Fix placing of xml declaration

ochaloup:
With current position of xml declaration the jta-crash-rec quickstart is failing to be deployed.
Putting the declaration on the start of the -ds.xml file.


sgilda:
@ochaloup : Good catch! This must have happened during the automated process to update license headers. I wonder how many more are in error? :-)

Thanks!

PR #497, opened 2013-05-02
FORGE-875: beans.xml is missing when run in JBoss Tools

gastaldi:
This fixes https://issues.jboss.org/browse/FORGE-875

Although the Faces scaffold already requires CDI, beans.xml is strangely not created in some installations. This ensures creation of beans.xml in the correct place.


rafabene:
PR merged

PR #498, opened 2013-05-09
Typo in the URL of AeroGear project site link is fixed.

duemir:
This is a quick fix using Github editor of the broken URL which I have noticed playing with JBoss Developer Studio.


sgilda:
This bug is also in the kitchensink-cordova/shared/www/index.html file. 
.

PR #500, opened 2013-05-09
JDF issue 243: Add jboss-cli scripts for all quickstarts that modify server configuration

sgilda:


LightGuard:
I haven't done anything with jboss-cli scripts, so I can't comment on those, but the rest looks fine.

sgilda:
I tested the scripts and they seem to work fine. Thanks for the review!

ejb-security-interceptors/README.md
```diff
@@ -111,28 +142,30 @@ After the server is configured you will then need to define four user accounts,
 
 This block of commands added a new security realm that is used by the quickstart, the most important part is the addition of the DelegationLoginModule which can be before or after the Remoting login module.
 
-	[standalone@localhost:9999 /] ./core-service=management/security-realm=ejb-outbound-realm:add
-	[standalone@localhost:9999 /] ./core-service=management/security-realm=ejb-outbound-realm/server-identity=secret:add(value="Q29ubmVjdGlvblBhc3N3b3JkMSE=")
+		[standalone@localhost:9999 /] ./core-service=management/security-realm=ejb-outbound-realm:add
+		[standalone@localhost:9999 /] ./core-service=management/security-realm=ejb-outbound-realm/server-identity=secret:add(value="Q29ubmVjdGlvblBhc3N3b3JkMSE=")
```
jamezp: Largely irrelevant, but the `./` isn't really needed from the root directory. Leaving it causes no real harm, but is inconsistent with the other scripts.

sgilda: Good catch. I'll change it to be consistent with the others.

sgilda:
@pmuir: The scripts were reviewed by @jamezp . Do you want to review these changes too?

jamezp:
Just thinking out loud here, but I wonder if it would be worth writing a simple maven plugin that would;
- Locate the CLI files
- Start Arquillian 
- Execute the commands

It would just give us a way to validate whether added scripts are valid. Probably the most useful for app server version changes.

I could probably write the plugin easy enough if anyone is interested.

pmuir:
Or would it be more generally useful to include this in Arqullian? I can see that the ability to alter the server config as part of the test would be pretty useful.

sgilda:
Arquillian sounds like the logical location to me. @jamezp : What do you think?
@pmuir : Is this OK to merge now?

jamezp:
I don't have a strong opinion either way. The one advantage I see to a maven plugin is that you can use it in the parent pom to locate the script files. I'm not sure how ARQ would work on a parent POM ,`<packaging>pom</packaging>`, project. One other advantage is it's not tied to the ARQ release cycle which also seems tied to the WildFly release cycle ATM.

Of course the disadvantage is we have another project to maintain. That said I started creating a maven plugin to replace expressions in properties file for WildFly. Maybe we could make that more generic and have a validation goal.

I did just find https://github.com/arquillian/arquillian-maven though. Maybe we could just resurrect that as it seems dead. I can take the lead on it if it's something we want to do. 

Anyway, this PR might not the best place for this discussion :)

sgilda:
@jamezp : Right. It will get lost here. How about the jdf-dev list?

sgilda:
This has been reviewed by @LightGuard and @jamezp .

@pmuir : Is it OK to merge?

PR #503, opened 2013-05-16
Fix compilation error in deltaspike quickstart

jaikiran:
The commit here fixes a compilation error in deltaspike quickstart


sgilda:
Thanks Jaikiran. Good catch! It appears I introduced this problem in a few other quickstarts also. I'll take care of them.

PR #506, opened 2013-05-22
JDF-327 updated to add in the jacorb config that is required for jts

tomjenkinson:


sgilda:
I will make the fixes.

PR #508, opened 2013-05-23
Update PicketLink quickstarts to 2.5.0.Beta3

sbryzak:


sgilda:
@sbryzak : I reviewed this early today and made comments, but apparently I didn't hit save.

Is there a reason these quickstarts are moving from version 1.0.6.Final to 1.0.7.CR2 of the JBoss BOM and the others are not? Should all of the other quickstarts be upgraded? We try to keep them in sync.

Also, when I run the QS Tools utility, some of the Java files are missing license headers.

sbryzak:
I've fixed the QS violations and pushed another commit.  The reason these quickstarts are using 1.0.7.CR2 is because they require PicketLink 2.5.0.Beta3 and won't work with previous releases. I'm not sure what the procedure is for updating the BOM for the other quickstarts but Pete will know.

sgilda:
Thanks @sbryzak . I'll investigate.

sgilda:
@sbryzak : Would you mind squashing this down to 1 commit? :-)

sgilda:
@sbryzak : Or, do you object if I squash it before I merge? ;-)

sgilda:
Squashed and merged.

PR #512, opened 2013-05-29
Fix quickstart issue 457: update commented EAP property versions

sgilda:
Modified the POM files so users can comment/uncomment to run with EAP 6.1.
Modified the base README file to provide solution to OutOfMemory error when running install from the base directory.

### DIFF
--- a/bean-validation/pom.xml
+++ b/bean-validation/pom.xml
@@ -48,10 +48,10 @@
         <!-- Define the version of the JBoss BOMs we want to import to specify tested stacks. -->
         <version.jboss.bom>1.0.6.Final</version.jboss.bom>
         <!-- Alternatively, comment out the above line, and un-comment the 
-            line below to use version 1.0.4.Final-redhat-3 which is a release certified 
+            line below to use version 1.0.4.Final-redhat-4 which is a release certified 
             to work with JBoss EAP 6. It requires you have access to the JBoss EAP 6 
             maven repository. -->
-        <!-- <version.jboss.bom>1.0.6.Final-redhat-3</version.jboss.bom> -->
+        <!-- <version.jboss.bom>1.0.4.Final-redhat-4</version.jboss.bom> -->
 
         <!-- other plugin versions -->
         <version.surefire.plugin>2.10</version.surefire.plugin>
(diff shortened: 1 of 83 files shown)


bean-validation/pom.xml
```diff
@@ -48,10 +48,10 @@
         <!-- Define the version of the JBoss BOMs we want to import to specify tested stacks. -->
         <version.jboss.bom>1.0.6.Final</version.jboss.bom>
         <!-- Alternatively, comment out the above line, and un-comment the 
-            line below to use version 1.0.4.Final-redhat-3 which is a release certified 
+            line below to use version 1.0.4.Final-redhat-4 which is a release certified 
             to work with JBoss EAP 6. It requires you have access to the JBoss EAP 6 
             maven repository. -->
-        <!-- <version.jboss.bom>1.0.6.Final-redhat-3</version.jboss.bom> -->
+        <!-- <version.jboss.bom>1.0.4.Final-redhat-4</version.jboss.bom> -->
```
LightGuard, line 54: Is this the correct version? I'd think it would be 1.0.6.Final-redhat-4 instead of 1.0.4.Final-redhat-4

sgilda:
@LightGuard : Yes. that's the version used here: https://github.com/jboss-eap/quickstart/blob/master-eap6/bean-validation/pom.xml

sgilda:
@LightGuard : I should mention that the 1.0.6.Final-redhat-4 was probably a result of a "change all" in the community POM files. EAP 6 never got to any 1.0.6 version.

LightGuard:
Ah, well that makes sense then.

On Wed, May 29, 2013 at 6:39 PM, sgilda notifications@github.com wrote:

> @LightGuard https://github.com/LightGuard : I should mention that the
> 1.0.6.Final-redhat-4 was probably a result of a "change all" in the
> community POM files. EAP 6 never got to any 1.0.6 version.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/jboss-jdf/jboss-as-quickstart/pull/512#issuecomment-18655164
> .

## 

Jason Porter
http://en.gravatar.com/lightguardjp

PR #528, opened 2013-06-13
Kitchensink-Cordova for iOS - remove useSplashScreen deprecated setting

tolis-e:
Removed the useSplashScreen deprecated setting

### DIFF
--- a/kitchensink-cordova/ios/KitchensinkCordova/Classes/AppDelegate.m
+++ b/kitchensink-cordova/ios/KitchensinkCordova/Classes/AppDelegate.m
@@ -84,7 +84,7 @@
 #else
         self.viewController = [[[MainViewController alloc] init] autorelease];
 #endif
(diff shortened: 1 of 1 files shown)


tolis-e:
@sgilda I think that an explanation would be useful. This line produces a warning in XCode while building the project and is not needed in Cordova versions >= 2.5.0. The only action needed in these newer versions of Cordova is to add or remove the SplashScreen plugin in config.xml. I thought that it would be better to comment this line and not to remove it for backward compatibility reasons.

sgilda:
@tolis-e : Sounds good. Let me know when you've added the comment and you need me to review again.

tolis-e:
@sgilda Comment added

VineetReynolds:
Squashed and merged: https://github.com/jboss-jdf/jboss-as-quickstart/commit/5f50a73aad92e929e5f9133ae9f8c4fb65111e00

PR #535, opened 2013-06-18
JDF-232: Clarify that most community artifacts are available in Maven Central or the JBoss Public Repository

sgilda:


sgilda:
@rafabene : I believe it is already there: https://github.com/jboss-jdf/jboss-as-quickstart/blob/master/README.md#configure-maven-for-jboss-enterprise-application-platform-6

rafabene:
When saying about JBoss Public Repository I had understood about JBoss Community repos: https://repository.jboss.org/nexus/content/groups/public/ and not EAP repository.

By the way. I think we should also update the #configure-maven-for-jboss-enterprise-application-platform-6 section to use the online version instead of local: http://maven.repository.redhat.com/techpreview/all/

What do you think ?

sgilda:
Waiting for responses to comments in JDF-232

rafabene:
Sorry! I forgot that transitive dependencies has the repositories declaration inside its pom.xml files. So it makes sense that "no additional configuration is needed". For me, it is good to merge!

PR #543, opened 2013-06-27
Updating Security BOM version to 1.7.0.CR7.

pedroigor:
Updating PicketLink Quickstarts to use the latest Security BOM version.


sgilda:
@pedroigor : It looks like these commits are being made directly to this repository and are not the result of a merge. If that is the case, would you mind squashing the commits?

pedroigor:
Going to send another one ...

PR #553, opened 2013-07-03
Fix Bug 980488: Change POST annotations to GET. Update README file.

sgilda:


VineetReynolds:
Looks good for merging otherwise.

sgilda:
@VineetReynolds : No we don't generally add that. I only left it there because I wasn't sure why they had it. I agree. with you. I will remove it. 

I'm also not sure why the last line about copyright headers is there. Should I remove that also?

PR #558, opened 2013-07-09
Update all Quickstarts to use 1.0.7.CR8 bom (aligned to WFK 2.3)

rafabene:


sgilda:
This looks good. Would you like me to merge it?

sgilda:
Both the root pom.xml and the helloworld-singleton/pom.xml files still have <version.jboss.bom>1.0.6.Final</version.jboss.bom>

@rafabene : Should they be updated?

rafabene:
Hummm, That's  because helloworld-singleton declared the property and doesn't use a JDF bom.
I've update them manually and I'll open a Jira to fix this

sgilda:
Modified root pom.xml file and merged.

PR #561, opened 2013-07-17
Extending naming rules for package and artifact id

antoinesd:
Working on Fuse quick starts, that have JBoss Fuse as server target, I propose that we extend the naming rules in Guidelines to reflect the example targeted product.


antoinesd:
Wrong PR.

PR #563, opened 2013-07-18
Extending naming rules for package and artifact id version 2

antoinesd:
This new revision contains notes in the first PR.


sgilda:
@antoinesd : In the future, you don't need to issue a new pull request. When you do your force update, it will appear in the original pull. :-)

antoinesd:
yes I know, but I missed up trying to squash commits, so I had restart from fresh

sgilda:
Looks good. Thanks!

PR #564, opened 2013-07-19
JDF-413 Make JPP quickstarts available on JDF site

ppalaga:


sgilda:
This was merged on 2013-07-19.

PR #565, opened 2013-07-22
JDF-412 - Move JDG Quickstarts to git submodules

rafabene:


sgilda:
@rafabene : This looks good to me. I do have one question though.

In the root pom.xml file, you removed the individual jdg quickstart modules and replaced them with just the root directory for the new submodule:

```
    <module>jdg-quickstarts</module>
```

I noticed the richfaces-quickstart commit replaced them with complete paths:

```
    <module>richfaces-quickstarts/helloworld-rf</module>
    <module>richfaces-quickstarts/kitchensink-rf</module>
```

I assume it's because that repository does not have a root pom.xml file. I think the richfaces quickstart repository should add a root pom.xml and the richfaces-quickstarts individual quickstart modules should be replaced by one only naming the root directory? Do you agree?.

rafabene:
Well. To avoid different kinds of submodules (direct one vs root pom) we can add it to richfaces repo and open a PR for them.

rafabene:
Since this enhancement doesn't change the result for jdf or richfaces. I think this pom.xml addition could be done on another PR. Do you agree?

sgilda:
Definitely! I just wanted to ask you while it was fresh in my mind. :-)

This pull looks fine! Would you like me to merge it?

PR #566, opened 2013-07-22
JDF-410 - Move richfaces-validation to richfaces-quickstarts submodules

rafabene:


sgilda:
Never mind. It works by naming the top level only. This looks good. Merging now.

sgilda:
@rafabene : I decided to run 1 more test. 

When I do the root `mvn clean install '-Pdefault,!complex-dependencies' , the`helloworld-rf`and`kitchensink-rf` build, but the richfaces-validation quickstart does not.

Since there is no root pom in the richfaces repository, should we keep the richfaces-validation module, but qualify it with the directory name as we do with the other 2 richfaces quickstarts? For example: 

```
            <module>richfaces-quickstarts/richfaces-validation</module>
```

rafabene:
Yes. We should. I miss that.

Thanks for the review. I'll update the PR.

sgilda:
Merged.

rafabene:
Merged

PR #567, opened 2013-07-23
JDF-430: Part I - Use new pattern for name in POM files.

sgilda:
Still need to rename directories and modify sub-module quickstarts


sgilda:
Added spaces between quickstart and subproject name.

PR #573, opened 2013-07-30
Cheat sheet for the helloworld quickstart

sgilda:


maxandersen:
This seems to be using the _old_ style actions instead of commands and it is using hardcoded paths instead of the relative path functionallity we added in JBDS 7.

I'll add a few pointers in the commits referring to https://github.com/maxandersen/cheatsheet-helloworld which shows the new and better approach.

helloworld/.cheatsheet.xml
```diff
@@ -0,0 +1,88 @@
...
+            via contexts.
+            <br />
+            <br />
+            <b>Servlet 3.0</b>
+            <br />
+            Servlet 3.0 introduces @Servlet, @ServletFilter, @FilterMapping, and other annotations to ease development of
+            servlet applications.
+        </description>
+    </intro>
+    <item skip="true" title="The helloworld example in depth">
```
maxandersen: One of your first items should have skip="false" to make sure the currentProject can be found and intialized.

Example: https://github.com/maxandersen/cheatsheet-helloworld/blob/master/cheatsheet.xml#L11

helloworld/.cheatsheet.xml
```diff
@@ -0,0 +1,88 @@
...
+            <b>WEB-INF/</b>
+            , which can be found in the
+            <b>src/main/webapp</b>
+            directory of the example.
+            Notice that we don&apos;t need a
+            <b>web.xml</b>
+            file.
+            This example does not contain a JSF file. The user interface is provided by the servlet.
+        </description>
+    </item>
```
maxandersen, line 42: At the end of the non-skippable item should be a <command> invocation like the one at https://github.com/maxandersen/cheatsheet-helloworld/blob/master/cheatsheet.xml#L16

``` xml
<command
            required="true"
            returns="currentProject"
            serialization="org.jboss.tools.project.examples.cheatsheet.getProjectForCheatsheet"/>
```

This ensures $currentProject will resolve to the project name used in eclipse - making it possible for the users to rename or move the project around and the cheatsheet will still work.

This is especially important if used for archetypes where the project name is set by the user as opposed to quickstarts which are somewhat static.

helloworld/.cheatsheet.xml
```diff
@@ -0,0 +1,88 @@
...
+            Let&apos;s take a look at the servlet that handles the request.
+        </description>
+        <subitem
+            label="The @WebServlet annotation provides the URL pattern for this servlet. This annotation replaces the web.xml servlet-mapping entries that were required before Servlet 3.0. "
+            skip="true">
+            <action pluginId="org.jboss.tools.project.examples.cheatsheet" class="org.jboss.tools.project.examples.cheatsheet.actions.OpenFileInEditor"
+                param1="/jboss-as-helloworld/src/main/java/org/jboss/as/quickstarts/helloworld/HelloWorldServlet.java" param2="43" />
+        </subitem>
+        <subitem label="The HTML page header and footer are defined as static strings. " skip="true">
+            <action pluginId="org.jboss.tools.project.examples.cheatsheet" class="org.jboss.tools.project.examples.cheatsheet.actions.OpenFileInEditor"
```
maxandersen: Don't use <action>'s if it can be avoided - they are fragile in comparison to <command>'s where parameters has names instead of indexes.

Plus commands can be looked up and browsed via cheatsheet editor and provides way different kind of actions you can do.

maxandersen: This action is better done like shown at https://github.com/maxandersen/cheatsheet-helloworld/blob/master/cheatsheet.xml#L27

``` xml
      <command
            required="false"
            serialization="org.jboss.tools.project.examples.cheatsheet.openFileInEditor(path=/${currentProject}/src/main/java/org/jboss/as/quickstarts/helloworld/HelloWorldServlet.java,fromLine=48,toLine=48)"/>
```

Notice the use of named parameters instead of `param3`

maxandersen: and notice the use of `${currentProject}` to avoid hardcoding the project path.

sgilda:
@maxandersen  : Thanks for taking the time to review this! I used the New Cheat Sheet wizard to create it and then stole some of this from Snjezana's kitchensink example, which as you said, is old style. I will make the updates and resubmit. 
Thanks again!

helloworld/.cheatsheet.xml
sgilda: Yes. Excellent!

pmuir:
Content looks good to me.

sgilda:
Thanks Pete. I was hoping you would review the content for me.

sgilda:
Updated to include fixes mentioned by @maxandersen.
Please let me know if I missed anything.

sgilda:
Merged.

PR #576, opened 2013-08-02
kitchensink-angularjs fix: Change datasource names for application and tests

sgilda:
...so they do not conflict with those used in the kitchensink quickstart


VineetReynolds:
+1 but you may want to repeat this for kitchensink-backbone and kitchensink-html5-mobile that have the same problem. I haven't checked the Spring kitchensink quickstarts.

sgilda:
Thanks @VineetReynolds  :-(

PR #585, opened 2013-08-13
Minor wording changes to Git Submodule instructions

sgilda:


sgilda:
Good catch! I had the hiccups I guess. ;-)

Better?

rafabene:
Much better! Ready to commit!

PR #586, opened 2013-08-13
JDF-455 remove maven compiler to remove JBDS issues

pmuir:

PR #589, opened 2013-08-26
Add compiler config to DeltaSpike QS to fix the compiler's failures for 1.3 source

OndraZizka:


sgilda:
@OndraZizka : It generates a warning, not an error. It still builds and deploys. Is that what you see?

rsmeral:
IMO, the original state before jboss-jdf@c33b9b2 caused the least trouble. The warning in Eclipse seems inevitable.
But if we're to include the maven compiler plugin (I think we should), let's define it's versions through properties, as it was before. WDYT?

sgilda:
@pmuir: I will defer to you. This pull reverses your pull https://github.com/jboss-jdf/jboss-as-quickstart/pull/586. The commit and WFK pulls are noted above.

OndraZizka:
What I see is really a failure of compilation. The source with annotations (and maybe generics) simply fails to compile.  Working with the latest OpenJDK.

OndraZizka:
mvn clean install '-Pdefault,!complex-dependencies'
...

[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 9.695s
[INFO] Finished at: Mon Aug 26 17:10:28 CEST 2013
[INFO] Final Memory: 54M/598M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:2.0.2:compile (default-compile) on project jboss-as-deltaspike-beanmanagerprovider: Compilation failure: Compilation failure:
[ERROR] /home/ondra/work/QS/qs-git/deltaspike-beanmanagerprovider/src/main/java/org/jboss/as/quickstarts/deltaspike/beanmanagerprovider/model/OPERATION.java:[25,7] error: enums are not supported in -source 1.3
[ERROR] 
[ERROR](use -source 5 or higher to enable enums)
[ERROR] /home/ondra/work/QS/qs-git/deltaspike-beanmanagerprovider/src/main/java/org/jboss/as/quickstarts/deltaspike/beanmanagerprovider/model/Contact.java:[41,1] error: annotations are not supported in -source 1.3
[ERROR] 
[ERROR](use -source 5 or higher to enable annotations)
[ERROR] /home/ondra/work/QS/qs-git/deltaspike-beanmanagerprovider/src/main/java/org/jboss/as/quickstarts/deltaspike/beanmanagerprovider/persistence/ContactRepository.java:[39,1] error: annotations are not supported in -source 1.3
[ERROR] 
[ERROR](use -source 5 or higher to enable annotations)
[ERROR] /home/ondra/work/QS/qs-git/deltaspike-beanmanagerprovider/src/main/java/org/jboss/as/quickstarts/deltaspike/beanmanagerprovider/persistence/ContactRepository.java:[47,17] error: generics are not supported in -source 1.3
[ERROR] 
[ERROR](use -source 5 or higher to enable generics)
[ERROR] /home/ondra/work/QS/qs-git/deltaspike-beanmanagerprovider/src/main/java/org/jboss/as/quickstarts/deltaspike/beanmanagerprovider/model/AuditContact.java:[34,1] error: annotations are not supported in -source 1.3
[ERROR] 
[ERROR](use -source 5 or higher to enable annotations)
[ERROR] /home/ondra/work/QS/qs-git/deltaspike-beanmanagerprovider/src/main/java/org/jboss/as/quickstarts/deltaspike/beanmanagerprovider/util/Resources.java:[46,5] error: annotations are not supported in -source 1.3
[ERROR] 
[ERROR](use -source 5 or higher to enable annotations)
[ERROR] /home/ondra/work/QS/qs-git/deltaspike-beanmanagerprovider/src/main/java/org/jboss/as/quickstarts/deltaspike/beanmanagerprovider/persistence/AuditContactListener.java:[54,5] error: annotations are not supported in -source 1.3
[ERROR] 
[ERROR](use -source 5 or higher to enable annotations)
[ERROR] /home/ondra/work/QS/qs-git/deltaspike-beanmanagerprovider/src/main/java/org/jboss/as/quickstarts/deltaspike/beanmanagerprovider/persistence/AuditContactListener.java:[57,12] error: generics are not supported in -source 1.3
[ERROR] 
[ERROR](use -source 5 or higher to enable generics)
[ERROR] /home/ondra/work/QS/qs-git/deltaspike-beanmanagerprovider/src/main/java/org/jboss/as/quickstarts/deltaspike/beanmanagerprovider/persistence/AuditRepository.java:[38,1] error: annotations are not supported in -source 1.3
[ERROR] 
[ERROR](use -source 5 or higher to enable annotations)
[ERROR] /home/ondra/work/QS/qs-git/deltaspike-beanmanagerprovider/src/main/java/org/jboss/as/quickstarts/deltaspike/beanmanagerprovider/persistence/AuditRepository.java:[64,15] error: generics are not supported in -source 1.3
[ERROR] 
[ERROR](use -source 5 or higher to enable generics)
[ERROR] /home/ondra/work/QS/qs-git/deltaspike-beanmanagerprovider/src/main/java/org/jboss/as/quickstarts/deltaspike/beanmanagerprovider/controller/ContactController.java:[45,1] error: annotations are not supported in -source 1.3
[ERROR] 
[ERROR](use -source 5 or higher to enable annotations)
[ERROR] /home/ondra/work/QS/qs-git/deltaspike-beanmanagerprovider/src/main/java/org/jboss/as/quickstarts/deltaspike/beanmanagerprovider/controller/ContactController.java:[65,16] error: generics are not supported in -source 1.3
[ERROR] -> [Help 1]

rafabene:
If we revert this we should use the same pattern: use maven properties instead of hardcoded version.

Probably it's easier to revert https://github.com/jboss-jdf/jboss-as-quickstart/commit/c33b9b2fb96b415ec7ca1e3b83b056bccf8487c4 instead of creating a new commit from scratch.

sgilda:
+1. This pull will not pass the QS tools utility tests.

I will leave this open for @pmuir to decide.

OndraZizka:
Ok, I'll use properties, although, I don't see any point. It just makes the pom more complex without adding any advantage. Maybe for future bulk update, but that's not worth. my2c

sgilda:
The properties make maintenance MUCH easer across 120+ quickstarts for these reasons:
- The are located near the top of the file in a known location
- You can easily do a 'find and replace' on a specific 'property name / version' combination using SED. In this case, the element name is unique, but in most other cases it is simply a `<version>xxx</version>` buried in the file and there's no way to easily determine what the version belongs to in an automated way.
- The QS tools utility is written to verify property values and report violations on hard-coded versions in the file.
- Like Java source code, hard-coding values that need to be updated on a regular basis is not good practice. :-)

OndraZizka:
I dare to disagree. All these points are better for us, not for the end user. Besides that:
1) We don't need to access or check this particular value (source / target = 1.6)
2) We won't change source format more than once in ages. If we do, I volunteer to do that in case we leave it as non-property.
3) QS tools can use multi-line regex to match this.
4) See 2.

Note this is about the source / target = 1.6 only.
But even for the deps versions:
pom.xml was originally intended as a structure to be edited by tools, not manally.
So it's the IDE and Maven plugins which should update the versons.
But I agree that in case of deps, we can give this up for the sake of manageability.

pmuir:
@OndraZizka what is your environment where this fails in maven?

pmuir:
If we need to specify these things, the best is to revert my previous commit, but I'm not sure why we do. All the quickstarts use annotations, and we don't have this problem elsewhere.

pmuir:
Also, it works fine for me.

sgilda:
@OndraZizka : I am closing this one because I think it was replaced by pull 594: https://github.com/jboss-jdf/jboss-as-quickstart/pull/594

Please let me know if this is not correct.

PR #602, opened 2013-09-03
First round of changes for reorganization, including updates to GAV and QS tools fixes

sgilda:


sgilda:
@rafabene : ejb-multi-server requires that you configure Maven to point to the product version of the Maven repository. That's why it's in the 'complex-dependency' profile. I did the configuration and it builds successfully.

PR #608, opened 2013-09-09
Added remove-configuration.cli to revert install-domain.cli changes back to default

Dantheman720:
Will make changes to README as well, but wanted to get comments on this.

### DIFF
--- a/ejb-multi-server/README.md
+++ b/ejb-multi-server/README.md
@@ -51,12 +51,24 @@ Modify the CLI Scripts (if you are running JBoss Enterprise Application Platform
 
 The CLI scripts provided with this quickstart target JBoss Enterprise Application Platform 6.1. If you are running older versions of the server, JBoss Enterprise Application Platform 6.0 or JBoss AS 7.1, you must modify the scripts to work against these servers.
 
+**Modify the Install-Domain Script:**
+
 1. Open the `install-domain.cli` file located in the root of this quickstart folder for editing.
 2. Find the lines that contain the following text:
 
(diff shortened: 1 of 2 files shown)


sgilda:
@Dantheman720 : Thanks! I'll look at this later today.

sgilda:
After running the script and comparing the configuration files to the backed up versions, I see one minor difference in the host.xml file: 

 The "ApplicationtRealm" security-realm is missing this line: 

```
    `<local default-user="$local" allowed-users="*" />`
```

sgilda:
I see quite a  few differences in the domain.xml file (first line is after removal script, second line is original file):

For example, in the Default profile and ha profile, under the  <subsystem xmlns="urn:jboss:domain:ejb3:1.4"> subsystem, 
                <default-missing-method-permissions-deny-access value="false"/>
previously it was 
                <default-missing-method-permissions-deny-access value="true"/>

There are  entries like this in the restored file that aren't in the original:

```
    <distributed-cache name="dist" l1-lifespan="0" mode="ASYNC" batching="true">
                    <eviction strategy="LRU" max-entries="10000"/>
                    <file-store/>
                </distributed-cache>
```

Have you done a diff on the domain.xml with the backed up version?

I need to look at this closer to see if the differences are important or not.

sgilda:
Great job. I think with a few minor changes, this is very close!

sgilda:
Thanks @Dantheman720 ! Excellent job.
Merged!

Dantheman720:
Thanks!

PR #609, opened 2013-09-10
Fix JDF-484: Modify quickstart GAVs for the reorg

sgilda:


sgilda:
Hold off on this. Will be adding fixes for JDF-485.

PR #611, opened 2013-09-11
Updating PicketLink Quickstarts.

pedroigor:


rafabene:
I run QSTools on it and it miss:
- Change Quickstarts groupId
- Property should be 'version.jboss.bom.eap'  instead of 'version.jboss.eap.bom'
- move picketbox-ldap and org.apache.httpcomponents:httpcore to `<dependencyManagement/>`

Tip: use QSTools 1.2.2.Final

PR #615, opened 2013-09-13
Missing forward slash breaks MemberResourceRESTService#lookupMemberById

paulrobinson:
The problem is that if AngularJS calls MemberResourceRESTService#lookupMemberById via the service defined in services.js, it will use this url, for a member with id '0':

http://localhost:8080/jboss-as-kitchensink-angularjs/rest/members0

This is fixed by adding the forward slash. 

This issue only arrises if you try and build on the Quickstart, as the AngularJS controller never looks up an individual member.


sgilda:
@paulrobinson: Good catch. Thanks! :-)

PR #620, opened 2013-09-17
Removed WFK Quickstarts

rafabene:


sgilda:
Great job! This was  lot of work. :-)

I grepped the READMEs in the old directory for 'Product Target: EAP' and the new directory structure matches. As long as the submodules will eventually disappear, this looks good to me.

PR #627, opened 2013-09-24
Remove references to JBoss AS 7 from the quickstarts

sgilda:


cdi-injection/README.md
```diff
@@ -26,7 +26,7 @@ Configure Maven
 If you have not yet done so, you must [Configure Maven](../README.md#configure-maven) before testing the quickstarts.
 
 
-Start JBoss Enterprise Application Platform 6 or JBoss AS 7 with the Web Profile
+Start JBoss EAP 6.1 with the Web Profile
```
rafabene, line 29: What about to suppress the "with the Web Profile" phrase ? I don't think we use this term. Just use: default/standalone, standalone/default-ha, full, full-ha

sgilda:
Could this wait for another pull? I'd like to get this in for Paul Gier.

PR #642, opened 2013-10-02
README: add Product Versions metadata, remove versions in text, remove temp repo from settings.xml

sgilda:


rafabene:
I like the result. It's good to merge!

PR #643, opened 2013-10-03
BZ1011851: Update version.jboss.ejb.client to use product artifact

sgilda:


sgilda:
Thanks @rafabene : That's where I found that version. :-)

PR #647, opened 2013-10-07
Merge from master to 6.2.x for ER5

sgilda:
@pgier: I think this is good to go. If you append '&w=1" to the URL when you compare, you can ignore whitespace differences that were fixed to comply with formatting requirements.

These are mostly README file changes to remove references to AS 7. We also removed most EAP version numbers throughout the text and refer only to 'EAP 6.1 or greater' near the beginning of the files.

Let me know if you see any problems.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Introduction
 ------------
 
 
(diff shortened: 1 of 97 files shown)


sgilda:
@pgier reviewed and commented on IRC that this is fine, so I will merge.

PR #651, opened 2013-10-08
BZ1015953: remove filename from Maven plugin. Mofify README deploy instructions for EARs

sgilda:


README.md
```diff
@@ -221,7 +221,8 @@ In some cases, you may want to build the application to test for compile errors
 2. Open a command line and navigate to the root directory of the quickstart you want to run.
 3. Use this command to build and deploy the archive:
 
-            mvn clean package jboss-as:deploy
```
rafabene, line 224: I think we can just leave 'mvn clean install jboss-as:deploy' as general instruction (install also packages). This will avoid a per packaging instruction

rafabene:
Other than general readme change, it looks good!

README.md
sgilda, line 224: I feel better stating the difference here. Otherwise, they may not realize it's different.

PR #658, opened 2013-10-09
BZ#1015948 Fixed test failures.

pgier:
Adopted use of @InSequence annotation to order tests for TaskDao.

Converted the Resources bean in tasks-jsf to a @RequestScoped one
to ensure that Arquillian injects a valid EntityManager instance
in the test instances.


sgilda:
if the merge of https://github.com/jboss-developer/jboss-eap-quickstarts/pull/657 tests out OK, this pull is not needed.

sgilda:
This commit was part of the bigger merge. Closing it because it's not needed.

PR #667, opened 2013-10-10
BZ1017762 fix for Logging on EAP 6.2.ER5

rafabene:


pmuir:
This is NOT the right approach. We need to understand why these changes are needed, and evaluate whether they break backwards compatibility, which on the face of it, they do.

rafabene:
This comment contains the explanation about why these changes were needed: https://bugzilla.redhat.com/show_bug.cgi?id=1017762#c3

PR #676, opened 2013-10-11
Bz1018311: Remove ejb-security-plus and ejb-security-propagation quickstarts

sgilda:
They use a private API


sgilda:
Reviewed by @pgier  and given the ok to merge.

PR #678, opened 2013-10-11
Bz1018269: Revert to private API and add a warning about the use of a private API to ejb-security-interceptors

sgilda:


sgilda:
Reviewed and approved by @pgier

PR #681, opened 2013-10-16
[BZ-1018311] Remove ejb-security-* quickstarts from modules

dosoudil:
Removing both quickstarts ejb-security-plus and ejb-security-propagation from maven modules declaration in Quickstarts Parent pom.xml. Fixes the issue from commit eda2795 which brokes compilation.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1018311


sgilda:
Merged.

PR #683, opened 2013-10-16
BZ#1011975 Adding picketlink-sts.war as quickstart.

sgilda:


picketlink-sts/src/main/webapp/WEB-INF/jboss-web.xml
```diff
@@ -0,0 +1,22 @@
...
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,  
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ -->
+
+<jboss-web>
+  <security-domain>picketlink-sts</security-domain>
+  <context-root>jboss-picketlink-sts</context-root>
```
rafabene: Do we need to force this context root ?

rafabene:
Except for the context-root, it seems fine.

PR #702, opened 2013-10-26
Fix README HTML styling. Copy CSS locally because it seems GitHub no longer serves up CSS.

sgilda:
Created a new directory in the dist folder: dist/css
Copied the stylesheets from @pmuir's github:
- https://github.com/pmuir/github-flavored-markdown/blob/gh-pages/shared/css/documentation.css
- https://github.com/pmuir/github-flavored-markdown/blob/gh-pages/shared/css/pygments.css

### DIFF
--- a/dist/css/documentation.css
+++ b/dist/css/documentation.css
@@ -1,0 +1,316 @@
+/*------------------------------------------------------------------------------
+     Global Documentation Styles
+------------------------------------------------------------------------------*/
+
+html {
+  height:100%;
+}
+
+body, table {
+  font: 13px helvetica,arial,freesans,clean,sans-serif;
+  line-height: 1.4em;
+  background-color: #fff;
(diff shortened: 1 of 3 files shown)


sgilda:
After working with @joshuaw to debug this on Friday, it appears the styling problem is not due to the disappearance of this file from GitHub:  https://raw.github.com/github/github-flavored-markdown/gh-pages/shared/css/pygments.css

I replaced that line with a file on @pmuir GitHub:  https://github.com/pmuir/github-flavored-markdown/tree/gh-pages/shared/css. However, that didn't solve the problem either. 

I created the files on my local file system and it styled correctly, so I added the files to the `dist/css/` directory structure so they will be part of the download and render the files correctly.

sgilda:
Tested by @joshuaw and it works. Merging the change.

PR #706, opened 2013-10-30
Bz1018311: Remove ejb-security-plus and ejb-security-propagation quickstarts from master - they use a private API

sgilda:


sgilda:
Merged.

PR #708, opened 2013-10-31
Add a contributor-settings.xml file, preconfigured with the temporary repository and 

sgilda:
update the CONTRIBUTING.md instructions


VineetReynolds:
Looks good otherwise. This would need to be added to wfk-quickstarts as well.

sgilda:
Changed the phrasing so it's more obvious that you choose between two methods. :-)

Once this is merged, I'll apply it to wfk and sandbox too.

PR #715, opened 2013-11-01
Removed colons from document title for guides.

VineetReynolds:
Colons were parsed resulted in some additional unwanted characters
being generated for the titles of these guides in the JDF site.


sgilda:
I believe you. Merged!

PR #721, opened 2013-11-05
Fix minor issues in CONTRIBUTING

sgilda:


rafabene:
It can be applied to both branches

sgilda:
Thanks @rafabene . I will take care of it!

PR #732, opened 2013-11-08
Remove the servlet-security-genericheader-auth from the 6.2.x branch since it uses private APIs

sgilda:


sgilda:
@pgier : I believe we want to remove this from 6.2, correct?

PR #750, opened 2013-11-12
Fiz BZ1029353 - Errors in helloworld-mbean quickstart

rafabene:
If it is ok, it should be applied to 6.2.x also


sgilda:
Added minor modification to readme and merged

PR #780, opened 2013-11-18
[BZ#1030937] Adding proper testing example to check if PicketLink STS works fine.

pskopek:
I didn't updated README.html as didn't know how it is generated.


sgilda:
I do not plan to merge this one. I will cherry pick from the master once it's working.

PR #790, opened 2013-11-21
Bz1025766: Change JBoss AS to JBoss EAP in pom file comments

sgilda:


sgilda:
@pgier : This fix replaces 'JBoss AS' with 'JBoss EAP' in the comments in the POM files. It may be too risky to merge to 6.2.x at this point. I'll let you decide.  

I am not able to test because of the '6.2.0-build-SNAPSHOT'.

bean-validation/pom.xml
```diff
@@ -162,7 +162,7 @@
                     <artifactId>maven-surefire-plugin</artifactId>
                     <version>${version.surefire.plugin}</version>
                 </plugin>
-                <!-- The JBoss AS plugin deploys your war to a local JBoss AS container -->
+                <!-- The JBoss EAP plugin deploys your war to a local JBoss EAP container -->
                 <!-- To use, set the JBOSS_HOME environment variable and run:
```
pgier, line 166: This change seems a bit strange because the plugin is still called "jboss-as-maven-plugin".  I would prefer to keep this as "JBoss AS plugin" unless we actually change the name of the plugin.

sgilda, line 166: Yes. I know. But we can't change the plugin name. I was told to replace all "JBoss AS" with "JBoss EAP" since there is no "JBoss AS".

We also can't change the 'jboss-as:deploy'

sgilda:
This was merged.

PR #808, opened 2013-11-26
use maven property project.groupId instead of having hardcoded in dependencies

wfink:
avoid problems if the projects groupId is changed


sgilda:
@wfink , Is this change required for the EAP 6.2 release?

wfink:
Not really,
it prevent from mistakes if you change the project.groupId (which I did 
to transfer it to wildfly) and change the project from jboss.eap => wildfly.
In that case the dependencies are loaded from the wrong version.

But it has no side effects.

sgilda:
Merged to master.

PR #809, opened 2013-11-26
ejb-multi-server: use maven property project.groupId instead of having hardcoded in dependencies

sgilda:


sgilda:
@pgier : I won't merge this unless you would like me to.

sgilda:
Closing since this has now shipped.

PR #822, opened 2013-12-05
Missing ID tag in profile to work with the activeProfiles tag.

Gonthim:
Build was failing if the temporary repo isn't active. It's in the <activeProfiles> tag, so I'm assuming the <id> tag in the profile was mistakenly left out. Also matched that profile's indentation with the others.


sgilda:
Good catch. Thanks!

PR #823, opened 2013-12-09
JDF-510: change the use of the scoped-context to avoid a TooManyChannelE...

wfink:
...xception and use the Context global for this bean to prevent also from performance issues due to massive re-creation and parallel connections for several bean instances. To reflect the correct terms 'EJBClient34' is renamed to to 'ScopedContext'.

Fix for https://issues.jboss.org/browse/JDF-510


ejb-multi-server/README.md
```diff
@@ -152,9 +152,9 @@ It also demonstrates how to invoke an EJB from a client using a scoped-context r
...
          
-    The line shows that the bean `MainEjbClient34App` is not secured and called at `app-main` server. The sub-calls to `app-one#` are using the scoped-context and the cluster view needs a time to be established. This is shown as the cluster-view call the `appOne` with the user `quickuser2`. `AppTwo` is called with two different scoped-context settings. Both are used alternately 7 times.
+    The line shows that the bean `MainAppSContext` is not secured and called at `app-main` server. The sub-calls to `app-one#` are using the scoped-context and the cluster view needs a time to be established. This is shown as the cluster-view call the `appOne` with the user `quickuser2`. `AppTwo` is called with two different scoped-context settings. Both are used alternately 7 times.
```
sgilda: I know this line didn't change much, but now that I read it again more closely, I'm not sure I understand what it is trying to say.

Should the first line be changed to the following? 

```
    The line shows that the bean `MainEjbClient34App` is not secured and called by `app-main` server.
```

Also, I find `app-one#` confusing because there are no numeric suffixes. Is this what it is trying to say?

```
    The sub-calls to `app-oneA` and `app-oneB` are using the scoped-context and the cluster view needs a time to be established.
```

I'm unclear which call the following statement refers to as there is no 'appOne' printed in the output anywhere.

```
   This is shown as the cluster-view call the `appOne` with the user `quickuser2`. 
```

sgilda:
@wfink : I tested this and it works great. If you can clarify the above, I will be glad to make the fixes to the README file and merge this pull.

wfink:
I've changed the explanation of the scoped-context result. Please check and let me know whether this is now more clear

ejb-multi-server/README.md
```diff
@@ -152,9 +152,15 @@ It also demonstrates how to invoke an EJB from a client using a scoped-context r
...
          
-    The line shows that the bean `MainEjbClient34App` is not secured and called at `app-main` server. The sub-calls to `app-one#` are using the scoped-context and the cluster view needs a time to be established. This is shown as the cluster-view call the `appOne` with the user `quickuser2`. `AppTwo` is called with two different scoped-context settings. Both are used alternately 7 times.
+    The resulting output in detail:
+    * shows that the bean `MainAppSContext` is called from the client without application security at the `app-main` server on host master.
```
sgilda: Now I understand what you're saying, but I may reword it:

 @wfink: Is the following correct or did I change the meaning?

```
* The client calls the `MainAppSContext` bean on the `app-main` server on host master with no application security.
* The `MainApp` bean in `app-main` calls the `AppOne` bean in `app-one` using the scoped-context and establishes the clustered view. 
    * It initially connects using `quickuser1` 
    * The clustered view is created using `quickuser2`. This takes some time, but once it takes effect, all calls are load-balanced.
* The calls to the 'AppTwo' bean in `app-two` are made using two different scoped-context settings and both are used alternately 7 times. This means the servers `app-twoA` and `app-twoB` are called alternately seven times each.
```

sgilda:
@wfink : Looks good. Thanks for making the changes.

@pmuir : Does this need code review? If yes, Jaikiran is no longer here, but Wolf suggested maybe dmlloyd or bmaxwell or wangc.

pmuir:
Good idea to get a review, yes.

sgilda:
@dmlloyd has agreed to review this pull request.

sgilda:
Good to go per @dmlloyd. Merging.

PR #844, opened 2014-02-14
Replace JDBS instructions with shared resource

sgilda:


rafabene:
It looks good!

sgilda:
Thanks again @rafabene

PR #853, opened 2014-02-20
JDF-647: FIx license headers for 2014

sgilda:


rafabene:
Huge PR but looks fine!

PR #855, opened 2014-02-20
Bz1067552: Fix helloworld-jms quickstart instructions to remove referenc...

sgilda:
...es to deployment and undeployment


rafabene:
Makes sense for me :)

PR #862, opened 2014-02-27
Bz1070761: Remove servlet-security-genericheader-auth from the build as it uses a private API

sgilda:


sgilda:
Reviewed and approved by @pgier

PR #896, opened 2014-03-18
Bz1077176: Update OpenShift instructions to document benign error message and fix environment variable name

sgilda:


rafabene:
I didn't know about this benign error but looks fine for me

PR #903, opened 2014-03-27
Bz1081446: replace JBOSS_HOME with EAP_HOME in ejb-multi-server and logging web app screens

sgilda:
Also add info about EAP_HOME variable to root README file.


rafabene:
It seems Ok. I had to read the _note_ to get sure that there wasn't a trailing JBOSS_HOME there.

PR #908, opened 2014-04-03
EAP6-71 remove dependencies to private JBoss API's

wfink:
The additional client and the SLSB are removed as they are not necessary for the example and use also private JBoss API's


wfink:
https://issues.jboss.org/browse/EAP6-71
https://bugzilla.redhat.com/show_bug.cgi?id=1081993

sgilda:
merged

PR #909, opened 2014-04-03
EAP6-71 remove dependencies to private JBoss API's

sgilda:
The additional client and the SLSB are removed as they are not necessary for the example and use also private JBoss API's


sgilda:
opened in error!

PR #948, opened 2014-05-09
Add command line arguments and fix wrong/missing comments

wfink:


sgilda:
Thanks Wolf. I'll try to review this soon.

sgilda:
This looks good! Is there a way I can test the remote server without an extra machine?

wfink:
You should have tested this before, right ;) ?
Anyway if you start the server with "-b <your IP address>" the default "localhost" will not work and you can use the approach with the properties file.

sgilda:
Yes, at the time, I probably should have, but it was crazy busy and I wanted to get it merge. If you have tested it and it works, I'm good. :-)

wfink:
I've had tested both :)

sgilda:
Merged.

PR #951, opened 2014-05-11
Fixed wrong port number

nak3:
Since it sets -Djboss.socket.binding.port-offset=100, the URL is 8180.


nak3:
Ah... Sorry! The line is for domain mode. Please DON'T pull it, it's my mistake.

sgilda:
@nak3 Thanks anyway for taking the time to try out the quickstart.

PR #959, opened 2014-05-22
JDF-706: Upgrade forge-from-scratch quickstart to use Forge 2

mbriskar:


sgilda:
DO NOT MERGE YET: This update is targeted for JBoss EAP 6.4!

I had a discussion with Lincoln, Max, and Matej. While you can use the command line, if you follow the instructions for JBoss Developer Studio, it requires version 8 beta. It also appears that Forge 2 may still be in a beta state.  

We  want to hold off on this update until after JBoss EAP 6.3 ships.

PR #982, opened 2014-06-09
Add a notice for appclient field injection

wfink:
Add a explanation and link to the JavaEE spec to clarify the injection context


sgilda:
@wfink : Hope you don't mind, but I made a minor wording change. It is now merged. Thanks!

PR #989, opened 2014-06-20
Add EmailAddress with UserType registered via package-info.java.

Ashutosh1490:
This works in AS 7.1.1 but fails with EAP 6.1, as the package-info.java
is not found.

Specifying the UserType directly via an @Type at the field works with both
jboss versions.

### DIFF
--- a/hibernate4/src/main/java/org/jboss/as/quickstart/hibernate4/model/Member.java
+++ b/hibernate4/src/main/java/org/jboss/as/quickstart/hibernate4/model/Member.java
@@ -29,9 +29,6 @@ import javax.validation.constraints.Pattern;
 import javax.validation.constraints.Size;
 import javax.xml.bind.annotation.XmlRootElement;
 
-import org.hibernate.validator.constraints.Email;
-import org.hibernate.validator.constraints.NotEmpty;
-
 /*The Model uses JPA Entity as well as Hibernate Validators
  * 
  */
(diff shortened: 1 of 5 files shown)


sgilda:
@Ashutosh1490 : I'm not clear what is not working. How do I recreate the problem?

The reason I ask is because it seems to be working for me and QA has passed it.

scottmarlow:
I'm not sure if its related but https://issues.jboss.org/browse/WFLY-1406 complains about Hibernate not being able to handle package-info in WildFly 8.0.0.Alpha2.  Its not clear from that jira if further work was done.

sgilda:
@pmuir, @rafabene, or @VineetReynolds:

Can you recommend someone to review the code changes for this pull as they are quite substantial. As I said above, the quickstart runs fine in JBoss EAP 6.3, so I'm not clear why the change is required.

hibernate4/src/main/java/org/jboss/as/quickstart/hibernate4/model/Member.java
```diff
@@ -53,9 +50,8 @@
 
     /** using hibernate4 validators **/
     @NotNull
-    @NotEmpty
-    @Email
-    private String email;
+    //@Type(type="org.jboss.as.quickstart.hibernate4.util.EmailAddressType")
```
rafabene, line 53: We should avoid having comments to remove a "dead" code

sgilda:
@rafabene : I think my basic questions is, do we need this update? And if so, why is it needed?

pmuir:
This is not the correct fix. It replaces something from _hibernate validator_ with a user type from hibernate.

Can you describe the problem in more detail?

PR #1150, opened 2014-11-26
Modify the interceptor to better handle local ejbs

dehort:


darranl:
Looks good

sgilda:
Thanks @dehort ! Merged!

sgilda:
This commit was reversed because some classes are not available in JBoss EAP.

PR #1182, opened 2015-01-01
Fix typo in comment for xa.btm

bparry02:


sgilda:
Thanks for the fix! Merged.

PR #1183, opened 2015-01-05
Bz1178644: Fix JBDS import warning for bmt quickstart

sgilda:


sgilda:
This was already merged.

PR #1219, opened 2015-01-07
Modify the interceptor to better handle local ejbs

sgilda:


sgilda:
@dehort, @darranl : This broke the build and had to be reverted.

PR #1224, opened 2015-01-08
Bz1178874: Fix JBDS import errors for tasks-jsf quickstart

sgilda:


sgilda:
This change is to fix https://bugzilla.redhat.com/show_bug.cgi?id=1178874

The fix needs code review before it can be merged!

paulrobinson:
This is what I do to get around this problem. However, I don't know if it's the preferred way to do it. You are probably best asking a CDI domain expert...

@antoinesd is this the right thing to do?

LightGuard:
It doesn't look like the TaskDao is being used up in the view area, so I would not go with `@Named`. Sure, `@Named` is a qualifier, but it's main purpose is to give it an EL name so you can use it in the view layer. I know it's a bit more work, especially for a fairly simple example, but I would suggest using a new qualifier. You could also go with `@Model`, which also gives it a name, but it gives it a bit more context (makes it `@RequestScoped` instead of `@Dependent`, you can see [this thread](https://developer.jboss.org/thread/179135?start=0&tstart=0) for some more info around that. If that doesn't really fit the intended use, create a new qualifier that more correctly matches the semantics you're looking for.

sgilda:
@LightGuard : This is beyond my skill level at this point. I need to find someone to help with the change you suggested.

sgilda:
Closing this pull as it is not the correct solution.

maxandersen:
any chance our validator is too agressive here ? 

the main problem though is that there is two - one for test and one for main.

sgilda:
@maxandersen : @LightGuard plans to fix this bug the correct way in another pull. I assume the validator is not too aggressive, but I'll let Jason comment on that. ;-)

maxandersen:
okey :)

PR #1233, opened 2015-01-08
Bz1178895: Fix JBDS import errors for the wsat-simple, wsba-coordinator-...

sgilda:
...completion-simple, and wsba-participant-completion quickstarts

### DIFF
--- a/helloworld-jms/README.md
+++ b/helloworld-jms/README.md
@@ -220,7 +220,7 @@ Run the Quickstart in JBoss Developer Studio or Eclipse
 -------------------------------------
 You can also start the server and deploy the quickstarts or run the Arquillian tests from Eclipse using JBoss tools. For more information, see [Use JBoss Developer Studio or Eclipse to Run the Quickstarts](https://github.com/jboss-developer/jboss-developer-shared-resources/blob/master/guides/USE_JBDS.md#use-jboss-developer-studio-or-eclipse-to-run-the-quickstarts) 
 
-This quickstart deploys and runs differently in JBoss Developer Studio than the other quickstarts.
+This quickstart consists of multiple projects, so it deploys and runs differently in JBoss Developer Studio than the other quickstarts.
 
 1. Configure and start the JBoss EAP server Red Hat JBoss Developer Studio:
    * Define a server runtime environment that uses the `standalone-full.xml` configuration file.


sgilda:
@paulrobinson : Would you like to review this pull before I merge it?

wsat-simple/src/main/webapp/WEB-INF/jboss-web.xml
```diff
@@ -15,7 +15,12 @@
...
     limitations under the License.
 -->
-<jboss-web>
+<!DOCTYPE jboss-web>
+<jboss-web xmlns="http://www.jboss.com/xml/ns/javaee"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
+    xsi:schemaLocation="http://www.jboss.org/schema/jbossas 
+    http://www.jboss.org/schema/jbossas/jboss-web_7_2.xsd">
+    <!-- Configure to enable WebSockets -->
     <context-root>/</context-root>
```
paulrobinson, line 19: Does this QS work with Web Sockets? I don't think we (the transactions team) have ever tried WS-AT or WS-BA with Web Sockets.

wsat-simple/src/test/java/org/jboss/as/quickstarts/wsat/simple/ClientStub.java
```diff
@@ -35,6 +37,6 @@
  */
 @Qualifier
 @Retention(RUNTIME)
-@Target({ TYPE, FIELD })
+@Target({TYPE, METHOD, FIELD, PARAMETER})
```
paulrobinson: You mentioned this was something you were not sure about. i think it's right:

TYPE: allows the injected class type to be annotated. 
METHOD: Used with @Produces on factory methods
FIELD: Typical use case of injecting into a class field
PARAMETER: One use would be to populate fields on a producer method.

wsba-participant-completion-simple/src/test/java/org/jboss/as/quickstarts/wsba/participantcompletion/simple/Client.java
```diff
@@ -57,6 +58,7 @@ public Client() throws MalformedURLException {
          * Add client handler chain so that XTS can add the transaction context to the SOAP messages.
          */
         BindingProvider bindingProvider = (BindingProvider) set;
+        @SuppressWarnings("rawtypes")
```
paulrobinson: I tried to fix this, so that the warning is never raised. However, setHandlerChain on line 64 doesn't seem to accept generic types. I'm not that hot on generics, but i think it could be fixed with some casting on line64, but it's probably not worth it.

paulrobinson:
I've finished reviewing. Generally fine, other than my concern about WebSockets.

wsat-simple/src/main/webapp/WEB-INF/jboss-web.xml
sgilda, line 19: Oops. Copied the comment from another file! Good catch!

wsba-participant-completion-simple/src/test/java/org/jboss/as/quickstarts/wsba/participantcompletion/simple/Client.java
sgilda: I tried too and gave up after a while. If you think of a solution later, feel free to update the code.

PR #1242, opened 2015-01-08
Fixes BZ1178874 (CDI warnings in tasks-jsf)

LightGuard:
I went with the alternative approach with a new Stereotype. This is
activated by a new beans.xml for testing. Unfortunately I had to do some
type casting because eclipse doesn't seem to know to look into / add /
whatever the test classes as possible injection points.


sgilda:
Thanks @LightGuard ! Merged.

PR #1323, opened 2015-01-21
Bz1182898: Attempt to fix JBDS tasks-rs runtime HTTP status 404 error

sgilda:


sgilda:
Do not merge! This does not work.

PR #1328, opened 2015-01-26
Removing 'org.jboss.jts' dependency

paulrobinson:


paulrobinson:
@sgilda this seems to work for me, but from what I can see it's the same as what you described doing in your email.

sgilda:
Thanks Paul! Merged.

PR #1330, opened 2015-01-28
Fix assorted errors reported by the maven-qstools-plugin

sgilda:


ejb-remote/pom.xml
```diff
@@ -20,7 +20,7 @@
     <modelVersion>4.0.0</modelVersion>
 
     <groupId>org.jboss.quickstarts.eap</groupId>
-    <artifactId>jboss-ejb-remote-parent</artifactId>
```
rafabene, line 23: Is this change from jboss-ejb-remote-parent to jboss-ejb-remote really expected ?

rafabene:
Cool! You catch a lot of things

ejb-remote/pom.xml
sgilda, line 23: @rafabene : qstools reported this as a problem because the artifactId doesn't match the folder name.

sgilda:
@rafabene : maven-qstools-plugin did it!

PR #1368, opened 2015-02-16
Only need to use external browser if broken

maxandersen:
Suggestion for stating that use of embedded browser is fine since it only affects OS's that have old webkit or xulrunner active in Eclipse


sgilda:
Thanks Max! I made a few minor edits and merged.

PR #1401, opened 2015-03-19
[BZ-1178748] Remove versions of security jars from pom

pgier:
These two versions are already set in the ejb-client-bom


sgilda:
@pgier: I don't know how to test quickstarts in the 6.4.x branch.

pgier:
I can just merge it and then test it in our Jenkins build
I tested the basic CLI build already by just modifying the previous build, so I think it would be ok.

sgilda:
@pgier : I applied this change to the ejb-security-interceptors quickstart pom.xml file from the CR1 build and tested that. It does fix the problem. I think it's good!

PR #1405, opened 2015-03-26
Fix QSTools violations for the quickstarts that were brought over from WFK

joshuawilson:
Fix QSTools violations for the quickstarts that were brought over from WFK.

I fixed the errors that I could but there are still some that I have no idea how to fix dealing with white space in the xml files.


joshuawilson:
@rafabene @sgilda This one is going to take some explaining. 

I noticed that all of my commits from the WFK repo had the JIRA reference stripped off.  I think this happened due to my custom of encasing the reference in square brackets [ ]. So the moved ones all just started with a dash - . I went back as far as I could and manually re-entered the JIRA reference numbers. But that required a massive rebase (108 commits) and a force push.  I don't know if you want this or not. I don't know of another way to fix it. Figured it would be better to clean it up now while we are still on a branch. ;)

Additionally I updated the xml files in the Spring Quickstarts to use the latest xml schema references since they were all pointing to an old version of Spring.

So in the list of commits, the last 2 have new code changes. The rest are just a rewording of the commit log message.

sgilda:
This one threw me a bit with all the commits. I think I'll let @rafabene review also since he's closer to these changes.

sgilda:
@rafabene : Do you think it would be safer to push this to a new branch name and preserve the original?

rafabene:
PR merged to moved-wfk-quickstarts branch

PR #1410, opened 2015-05-05
Update quickstarts for EAP 7/Wildfly 9

pgier:
There are three quickstarts which are not building currently due to dependency issues.
ejb-security-interceptors
servlet-security-genericheader-auth
cluster-ha-singleton

They are commented out for now, but need to be fixed in a separate commit

### DIFF
--- a/app-client/client-simple/pom.xml
+++ b/app-client/client-simple/pom.xml
@@ -22,7 +22,7 @@
     <parent>
         <groupId>org.jboss.quickstarts.eap</groupId>
         <artifactId>jboss-app-client</artifactId>
-        <version>6.4.0-redhat-SNAPSHOT</version>
+        <version>7.0.0-SNAPSHOT</version>
      </parent>
 
     <artifactId>jboss-app-client-client-simple</artifactId>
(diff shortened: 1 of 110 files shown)


sgilda:
@pgier  , is there a fix for those quickstarts pending or will they just not work in Wildfly 9? I believe they all used private APIs so I assume they will need total rework.

@rafabene : Should we move those 3 quickstarts to Sandbox for now, or should we go ahead and commit them and add a note that they do not work?

pgier:
@sgilda The quickstarts need more investigation I think regarding how much work they will need to be fixed.  I would probably just leave them in but commented out for this first DR.  For the next DR we can see whether they can be fixed relatively easily or need to be removed.

rafabene:
@sgilda Note that those 3 sandbox quickstarts that you mentioned are not listed on the plan.

The moving quickstarts from sandbox are:
- batch-processing
- jaxrs-client
- jsonp
- managedexecutorservicea
- websocket-endpoint.

sgilda:
@rafabene : That's why I think maybe we should move them to Sandbox until the APIs are worked out.

PR #1442, opened 2015-07-17
JDF-844: Add a profile for the Arquillian tests to allow the default …

sgilda:
…profile to skip all tests on the mvn install command.


sgilda:
@rafabene : This modification seems to fix the issue, but could you please verify that it is the correct fix?

rafabene:
Sounds that the fix is valid!

sgilda:
Thanks @rafabene !

PR #1445, opened 2015-07-27
Removed comment about non-existent arquillian tests from spring-reste…

trepel:
…asy pom.xml


sgilda:
Thanks @trepel : Merged!

PR #1448, opened 2015-07-27
Fixed renaming of contacts-mobile-basic example to contacts-jquerymobile

trepel:


sgilda:
When I couldn't build and access the quickstart for the previous pull. I made and applied these fixes. Sorry about that. Closing this pull because it is not needed.

PR #1460, opened 2015-08-05
Update JPA version to 2.1 in persistence.xml

trepel:
JPA 2.1 is part of Java EE 7 Specification, EAP 7 is Java EE 7 container, thus every quickstart targeting EAP7 should use JPA 2.1.


sgilda:
Merged!

PR #1461, opened 2015-08-05
Remove versions in Spring schema locations in spring-greeter example

trepel:
In Spring, recommended practice is not to specify versions in XSD files. In that case the latest version is used, which is perfectly ok, even if latest version is not used in application. Specific versions should be used only if application doesn't work with latest (unversioned) XSD files.


sgilda:
merged!

PR #1477, opened 2015-08-10
JBEAP-631: Fix warning - WFLYEJB0463: Invalid transaction attribute t…

sgilda:
…ype REQUIRED on SFSB lifecycle method initNewMember()


sgilda:
@rafabene : Does this look like the correct fix? I'm just guessing on this. :-)

PR #1484, opened 2015-08-13
Updated wro4j-maven-plugin to latest version 1.7.8

trepel:
Currently this is the latest version. Furthermore, this version is also used in kitchensink-html5-mobile example, I believe we should keep the versions aligned in these two quickstarts.


sgilda:
Merged

PR #1485, opened 2015-08-13
Remove jboss-deployment-structure.xml from contacts-jquerymobile

trepel:
EAP7 scans war during deployment, detects the usage of Jackson library, and enables Jackson modules automatically. Thus there is no need to declare the dependency on Jackson modules explicitly using jboss-deployment-structure.xml.

Moreover, the content of this file is incorrect because the com.fasterxml.jackson modules (Jackson2) should be used, not org.codehaus.jackson modules. These modules are private and are not supposed to be used by user applications directly.

This PR deletes the file because I think quickstarts should be kept as simple as possible.


rafabene:
@sgilda  Can you merge this PR, please ? Thanks

sgilda:
@rafabene or @trepel : did you test this to make sure it works?

sgilda:
Merged!

rafabene:
I didn't test it but I suppose that @trepel did.

sgilda:
@rafabene : I did. :-)

PR #1492, opened 2015-08-14
Update Spring-Petclinic 

joshuawilson:
This will change the pom and few files to match the Pivitol version of Spring-Petclinic.


spring-petclinic/pom.xml
```diff
@@ -108,83 +110,75 @@
...
+		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
 		<!-- JBoss dependency versions -->
-		<jboss.maven.plugin.version>7.4.Final</jboss.maven.plugin.version>
+		<version.wildfly.maven.plugin>1.0.2.Final</version.wildfly.maven.plugin>
         <version.jboss.bom.eap>7.0.0-build-7</version.jboss.bom.eap>
+		<version.jboss.spec.javaee.7.0>1.0.3.Final</version.jboss.spec.javaee.7.0>
+
+		<!-- Generic properties -->
+		<java.version>1.7</java.version>
```
rafabene: This seems to be an unused property and conflicts with the definition about 1.8 bellow. Can we remove it or you just left it to be closed to the original version ?

spring-petclinic/pom.xml
```diff
@@ -195,13 +189,19 @@
                  read this as the JBoss stack of the Java EE APIs). You can actually
                  use this stack with any version of JBoss EAP that implements Java EE. -->
             <dependency>
+                <groupId>org.jboss.spec</groupId>
```
rafabene: I think we can remove this BOM since jboss-javaee-7.0-eap-with-tools includes it

spring-petclinic/pom.xml
```diff
@@ -626,10 +352,7 @@
 		</dependency>
 
 		<!-- For MySql only -->
-		<!-- <dependency>
-			<groupId>mysql</groupId>
-			<artifactId>mysql-connector-java</artifactId>
-		</dependency> -->
+		<!-- <dependency> <groupId>mysql</groupId> <artifactId>mysql-connector-java</artifactId> </dependency> -->
```
rafabene: This lost the format. I assume that it's safe to remove it.

spring-petclinic/pom.xml
```diff
@@ -748,21 +443,21 @@
 			<plugin>
```
rafabene, line 424: As per my understanding we can remove the whole maven-compiler-plugin.

joshuawilson, line 424: I am not sure what to do about this. The Pivitol version has this plugin. I only set it to how we do it. I can leave the whole thing running on Java 7 and then it would really match upstream. What do you suggest?

spring-petclinic/pom.xml
joshuawilson: That is the way they have it in the upstream. I would remove it but I think it is there in case you want to use MySQL. I can fix the formatting but it won't match then. I can do either, what do you prefer?

rafabene: In case of mysql I think it will be handled by EAP7 and not by the application. The application doesn't deal directly with MySQL. So I assume it's safe to remove the mysql stuff from the pom.xml

spring-petclinic/pom.xml
rafabene, line 424: Based on what we talked and the purpose of the quickstart ( "is to show how to take a Spring app and deploy it on EAP/WildFly") we can also put some of our flavours on it. This means that we can remove the maven-compiler-plugin and use the maven.compiler.source and maven.compiler.target properties to make it aligned with Java SE 8. This is how we do it for the other quickstarts.

sgilda:
@rafabene , @joshuawilson : Are we still waiting for updates to this pull?

joshuawilson:
Yes, I am working through the QSTools findings now.

joshuawilson:
I updated the pom with the above comments and ran qs-tools.

sgilda:
Testing this, I get an error:

17:17:32,492 WARN  [warn](default task-45) Handler execution resulted in exception: java.lang.RuntimeException: Expected: controller used to showcase what happens when an exception is thrown
    at org.springframework.samples.petclinic.web.CrashController.triggerException(CrashController.java:35)

sgilda:
Merged!

PR #1526, opened 2015-08-25
JBEAP-677  ejb-remote; ejb-security-interceptors; ejb-asychronous -- correct maven dependencies for client applicaton

wfink:
https://issues.jboss.org/browse/JBEAP-677

maven dependencies are not set correct, so the client is missing some libraries


sgilda:
Thanks Wolf! They work now. :-)

PR #1545, opened 2015-08-31
JBEAP-911: Update JMS code in cmt quickstart

sgilda:


sgilda:
@rafabene : I only updated the JMS and JPA code and instructions for this pull. I left the rest of the logging and error handling as it was. Can you please review again? Thanks!

rafabene:
This change makes more sense for me now. But  I think we should also remove ths quickstart from requires-postgres maven profile at the root pom.xml

It's much better now that the JMS API was updated too.

sgilda:
Removed the cmt quickstart from the PostgreSQL restriction in the root pom.xml.

rafabene:
Sorry, I should say `move` from requires-postgres to default profile in the root pom.xml

sgilda:
@rafabene : Good catch! I think I'm too scattered to focus these days. ;-)

PR #1577, opened 2015-09-09
Added the "exclude-datasource" profile to allow using a managed datasource

luksa:
We need (something like) this at the CE team in order to use the kitchensink quickstart with datasources configured by our OS3 templates.

### DIFF
--- a/kitchensink/pom.xml
+++ b/kitchensink/pom.xml
@@ -294,5 +294,21 @@
             </build>
         </profile>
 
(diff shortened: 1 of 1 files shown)


luksa:
Closing this, because we actually need the war to not include the datasource even when you don't activate any additional profiles.

I solved this by creating a new project that uses maven-war-plugin overlays to remove the file.

PR #1583, opened 2015-09-11
spring-greeter dependency cleanup

trepel:
The `taglib:standad` dependency must not be used. It contains JSTL, we use `jboss-jstl-api_1.2_spec` shipped with EAP instead.
The `commons-logging` and `aopalliance` dependencies don't have to be specified, they are transitively fetched anyway.
Java EE 7 specifications are shipped with EAP, thus their scope should be `provided`


sgilda:
Thanks for all the fixes. 
This has been merged.

PR #1584, opened 2015-09-11
spring-kitchensink-basic dependency cleanup

trepel:
The `taglib:standad` dependency must not be used. It contains JSTL, we use `jboss-jstl-api_1.2_spec` shipped with EAP instead.
The `commons-logging` and `aopalliance` dependencies don't have to be specified, they are transitively fetched anyway.
The `cglib` dependency is not required as of Spring 3.2, see http://docs.spring.io/spring/docs/3.2.x/spring-framework-reference/html/migration-3.2.html
The `jackson` dependencies don't need to be explicitely specified, they are automatically provided by EAP7 as static modules
Java EE 7 specifications are shipped with EAP, thus their scope should be `provided`
The `slf4j-simple` dependency is not used by the example, can be safely removed.

### DIFF
--- a/spring-kitchensink-basic/functional-tests/pom.xml
+++ b/spring-kitchensink-basic/functional-tests/pom.xml
@@ -176,7 +176,7 @@
         <dependency>
             <groupId>org.json</groupId>
             <artifactId>json</artifactId>
-            <version>20090211</version>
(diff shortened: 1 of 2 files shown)


sgilda:
Thanks again! Merged.

PR #1589, opened 2015-09-15
spring-kitchensink-springmvctest dependency cleanup

trepel:
The taglib:standad dependency must not be used. It contains JSTL, we use jboss-jstl-api_1.2_spec shipped with EAP instead.
The commons-logging and aopalliance dependencies don't have to be specified, they are transitively fetched anyway.
The cglib dependency is not required as of Spring 3.2, see http://docs.spring.io/spring/docs/3.2.x/spring-framework-reference/html/migration-3.2.html
The jackson dependencies have to be explicitly specified for testing only, version updated according to the one shipped with WildFly 10.
Java EE 7 specifications are shipped with EAP, thus their scope should be provided.
The slf4j-simple dependency is not used by the example, can be safely removed.
The `json` and `json-path` updated to latest versions

### DIFF
--- a/spring-kitchensink-springmvctest/functional-tests/pom.xml
+++ b/spring-kitchensink-springmvctest/functional-tests/pom.xml
@@ -176,7 +176,7 @@
         <dependency>
             <groupId>org.json</groupId>
             <artifactId>json</artifactId>
-            <version>20090211</version>
+            <version>20141113</version>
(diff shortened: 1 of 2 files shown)


sgilda:
@trepel : Thanks so much for cleaning up the Spring quickstarts!

Merged.

PR #1636, opened 2015-10-12
Update of maven-war-plugin, jQuery and small fixes of README.md in helloworld-html5

trepel:


helloworld-html5/README.md
```diff
@@ -63,8 +63,8 @@ The application will be running at the following URL <http://localhost:8080/jbos
 
 You can also test the REST endpoint as follows. Feel free to replace `YOUR_NAME` with a name of your choosing.
 
-* The *XML* content can be tested by accessing the following URL: <http://localhost:8080/jboss-helloworld-html5/hello/xml/YOUR_NAME> 
-* The *JSON* content can be tested by accessing this URL: <http://localhost:8080/jboss-helloworld-html5/hello/json/YOUR_NAME>
+* The *XML* content can be tested by accessing the following URL using HTTP POST: <http://localhost:8080/jboss-helloworld-html5/hello/xml/YOUR_NAME> 
+* The *JSON* content can be tested by accessing this URL using HTTP POST: <http://localhost:8080/jboss-helloworld-html5/hello/json/YOUR_NAME>
```
sgilda, line 67: @trepel : All of the quickstarts assume the they should use a browser to access the application URLs. While it uses HTTP POST, I'm afraid this language might confuse some users. They may think they need to use a CURL command. 

I wonder if we should make it clear that we intend them to access the URLs in a browser. WDYT?

trepel, line 67: @sgilda: I was reading the README.md and when I hit that lines, I just copy/paste the URLs to the browser. It resulted in error and I thought that something was wrong with the quickstart. But after looking into source code I realized that HTTP POST had to be used. I believe that the current wording is confusing. Since browsers use HTTP GET user must use some other tool or install browser plugin to be able to use that URLs for REST endpoint testing.

sgilda:
@trepel : Yes, you are right. I didn't look closely enough. Sorry about that.

In that case we should do something like what was done in the tasks-rs quickstart [1] README and provide the 'curl' command example.

Do you want to add that or would you prefer I did it?

[1] https://github.com/jboss-developer/jboss-eap-quickstarts/tree/7.0.x-develop/tasks-rs#create-a-task

trepel:
@sgilda : Thank you for the link. I think it will be better if you do the wording. Nevertheless I created https://github.com/jboss-developer/jboss-eap-quickstarts/pull/1645 with my proposal. You might find it useful.

sgilda:
@trepel : One last favor. Would you mind rebasing and squashing the commits? Thanks!

sgilda:
Squashed and merged.

PR #1637, opened 2015-10-12
JBEAP-1122 Use default for use-ccm

mmusgrov:
https://issues.jboss.org/browse/JBEAP-1122

The server startup message is saying that lazy enlistment isn't available because use-ccm is set to false. This is is just a warning during deployment and the actual quickstart obtains the connection inside a transaction.
I have updated the datasource definition (and README) to force the use of the default setting for this attribute.


sgilda:
Thanks Mike! Works great. :-)

PR #1638, opened 2015-10-12
Functional test extended to cover empty field submission in helloworl…

trepel:
…d-html5


sgilda:
Lookd good. Thanks @trepel !

PR #1659, opened 2015-10-16
Updated resteasy client libs and updated JAX-RS classes

blainemincey:
Updated the RestEasy Client libraries and the associated JAX-RS libraries.  This required minor updates to the pom.xml as well as minor updates to the code within the test Java class.  This update is based upon https://issues.jboss.org/browse/JBEAP-1407 .


sgilda:
Worked great! Thanks @blainemincey !

PR #1662, opened 2015-10-19
Added Themes menu item to all menus in contacts-jquerymobile

trepel:
There is no reason why to have "Themes" menu item only in one of the menu, apart of this menu item all the menus are almost the same.

Together with "Themes" menu item addition also two typos were fixed and one redundant license comment was removed.

### DIFF
--- a/contacts-jquerymobile/src/main/webapp/index.html
+++ b/contacts-jquerymobile/src/main/webapp/index.html
@@ -15,22 +15,6 @@
     limitations under the License.
 -->
 <!DOCTYPE html>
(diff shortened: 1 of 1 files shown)


sgilda:
@trepel : Thanks!

joshuawilson:
The Theme stuff was only an experiment and might be best if it was removed
all together. I'll leave that up to you to decide.

trepel:
@joshuawilson : Thanks for information. It causes no problems so far, I believe we can keep it for now. I will take a look whether it is possible to test it in an automated way. If not, I will consider the option of removal the Theming.

PR #1668, opened 2015-10-21
JBEAP-1421: m2e mapping added to jaxws-retail

jbliznak:
Prepared PR with proposed fix from Fred Bricon https://issues.jboss.org/browse/JBEAP-1421


sgilda:
Thanks @jbliznak ! I started with a clean JBDS, clean quickstart, and this worked. Thanks!

PR #1669, opened 2015-10-21
Replaced deprecated annotation from Spring quickstarts

trepel:
@TransactionConfiguration has beed deprecated in Spring 4.2, in this particular case it  can be replaced with @Rollback


sgilda:
Merged!

PR #1680, opened 2015-10-26
Small fixes for qstools check to pass in kitchensink-html5-mobile

trepel:
Removal of unused property in pom.xml; moving DOCTYPE header after licence comment in two xml files.


sgilda:
Thanks @trepel : Merged!

PR #1687, opened 2015-10-27
JBEAP-1438: Update picketlink-sts JBDS instructions. Add default inde…

sgilda:
…x page for JBDS to display. Change context from picketlink-sts to jboss-picketlink-sts to be consistent with other quickstarts


sgilda:
@pskopek : Can you please review this change before I merge it?
JBDS expects deployment of a WAR to result in display of the application in a browser. WARs that have no UI result in a 404. We have resolved this issue in the past by creating an index.html welcome file. 

This quickstart required additional changes. The context root wasn't consistent with the other quickstarts, which have the 'jboss-' prefix: 'jboss-<quickstart-name>'. I'm pretty sure I caught all instances that needed changing, but would appreciate it if you could double-check for me.

Also, to display the welcome page, the user is challenged for the UserName and Password. I add that information to the JBDS instructions.

Let me know if I missed anything.

sgilda:
@pedroigor reviewed this and commented in #picketlink-sts IRC channel that it looks good.

PR #1688, opened 2015-11-03
Rebase spring-petclinic to match latest upstream version

trepel:
This rebase fixes all the issues discovered in spring-petclinic example, all functional tests should pass.


sgilda:
@trepel : I still see one warning when I access the application:

```
    WARN  [com.github.dandelion.core.config.StandardConfigurationLoader] (default task-1) No file "dandelion.properties" was found in "dandelion/dandelion.properties" (classpath). The default configuration will be used.
```

Is that expected?

trepel:
@sgilda: yes, this warning appears in origin too. They apparently don't mind and it has no effect on example. I think it is minor enough so we don't have to bother, also I think it is not necessary to have it in README either.

If you feel it differently, I can try to fix it. Just let me know.

sgilda:
@trepel : No, it's fine, I just need to add it to the README. I assume it looks for that file in case you have a custom configuration, but if not, uses the default?

trepel:
Yes, the example doesn't require any change of dandelion default configuration so they omitted the dandelion.properties completely.

Ok, if you feel, add it to README. I might try to get rid of this warning if I have time.

sgilda:
@trepel : I wouldn't worry about it. It's a useful warning. :-)

PR #1695, opened 2015-11-16
Fixed licenses and added angularjs quickstart to main pom.xml

trepel:
Mainly just change of the year in licences and also I forgot to add the quickstart into main pom.xml, so I fixed it by this PR.


sgilda:
@rafabene : I ran QSTools against this quickstart when I merged it and it did not report the license violations. Does QSTools need an update?

sgilda:
@trepel : Thanks for the fix!

PR #1696, opened 2015-11-18
Update of several dependencies, fixed name and description elements

trepel:
Fixed name and description elements in pom.xml files
Updated maven-war-plugin and maven-surefire-plugin
Update of several dependencies


sgilda:
Looks good. Thanks @trepel !

PR #1697, opened 2015-11-18
Minor fixes of arquillian.xml files

trepel:
Fixes of comments in arquillian.xml files - maily version removals. It is not required there and it is difficult to maintain.

Element engine must be placed after element defaultProtocol


sgilda:
Thanks @trepel . Merged with this commit: https://github.com/jboss-developer/jboss-eap-quickstarts/commit/f32f4c0a9f2cc58832a6393f06ebab03d59891d9

PR #1698, opened 2015-11-18
Minor fixes to make qstool maven plugin check happy

trepel:
Indentation issues, one summary fix - was too short, and specifying dependency version via dependencyManagement rather than having in in dependencies.


sgilda:
Merged with this commit: https://github.com/jboss-developer/jboss-eap-quickstarts/commit/6e8b1b3c2f38a537c6a50ed859e34320d1a56ccd

PR #1700, opened 2015-11-25
Int-tel-input update to 6.4.3 in contacts-jquerymobile

trepel:
Update separeted to two commits. In first there is just rename of files and updated paths to that files accordingly. In second commit there is update itself.

Also I added the license as required by the author, see https://github.com/jackocnr/intl-tel-input/blob/master/LICENSE


sgilda:
Thanks @trepel !

PR #1704, opened 2015-12-02
JBEAP-2075: Fix potential IndexOutOfBoundException in ejb-security-in…

sgilda:
…terceptors quickstart


sgilda:
@darranl : Can you please review? I tested the quickstart with the fix and it works.

ejb-security-interceptors/src/main/java/org/jboss/as/quickstarts/ejb_security_interceptors/DelegationLoginModule.java
```diff
@@ -140,7 +140,7 @@ protected boolean delegationAcceptable(String requestedUser, OuterUserCredential
         }
 
         String[] allowedMappings = loadPropertyValue(connectionUser.getName(), connectionUser.getRealm());
-        if (allowedMappings.length == 1 && "*".equals(allowedMappings[1])) {
+        if (allowedMappings.length == 1 && "*".equals(allowedMappings[0])) {
```
rafabene, line 143: This makes sense for me. If it's a 1 size array, than we should compare the first position of it. It's ok to merge.

sgilda:
@rafabene : Thanks again!

PR #1708, opened 2015-12-03
Removed hard-coded version in kitchensink-angularjs functional tests

trepel:
Version should be resolved using EAP7 -with-tools BOM.


sgilda:
Tested and it works fine. Thanks @trepel !

PR #1710, opened 2015-12-07
Updated of QUnit library to 1.20.0 in Kitchensink HTML5 Mobile

trepel:


sgilda:
Looks good. Thanks @trepel

PR #1714, opened 2015-12-15
Removal of redundant earlyaccess repository

trepel:
This "earlyaccess" repo is redundant, "earlyaccess/all" repo must be sufficient.

### DIFF
--- a/helloworld-html5/functional-tests/pom.xml
+++ b/helloworld-html5/functional-tests/pom.xml
@@ -43,16 +43,6 @@
     -->
     <repositories>
         <repository>
-            <id>redhat-earlyaccess-repository</id>
-            <url>https://maven.repository.redhat.com/earlyaccess/</url>
-            <releases>
-                <enabled>true</enabled>
-            </releases>
-            <snapshots>
-                <enabled>false</enabled>
-            </snapshots>
-        </repository>
-        <repository>
             <id>jboss-developer-staging-repository</id>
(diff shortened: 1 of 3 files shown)


sgilda:
@trepel : Good catch! How did you notice this?

It looks like these quickstarts already had an earlyaccess repo defined before @pgier did the replacement from ga --> earlyaccess with this commit [1], so they ended up with 2 instances.

https://github.com/jboss-developer/jboss-eap-quickstarts/commit/ab32494c807e8f233852146afe2ba4e2e86ed69c

Thanks!

trepel:
@sgilda I have noticed this when I was investigating the qstools complaints. I was trying to find out why qstools complains about earlyaccess/all repository, I found the two issues (one of them was https://issues.jboss.org/browse/JBEAP-1434). So I wanted to figure out whether the earlyaccess/all could be removed altogether, because we are after Beta already. And I discovered this duplicity.

Whether the earlyaccess/all can be removed or not I am not sure yet. But probably yes because it doesn't contain 7.0.0.Beta1 version, see for instance https://maven.repository.redhat.com/earlyaccess/all/org/jboss/bom/jboss-eap-javaee7-with-tools/

sgilda:
@trepel : I imagine @pgier is waiting for the go ahead from QA before tagging the quickstarts and changing them back to ga. I'll try to check with him today.

Thanks again for all your help!

PR #1715, opened 2015-12-15
Added JTA API dependency for tests to Spring Quickstarts

trepel:
Hibernate 5.0.5 no longer transitively depends on JTA API so it needs to be added explicitly.


sgilda:
Thanks @trepel .

PR #1723, opened 2016-01-07
JBEAP-2598: Update Bootstrap CSS link in payment-cdi-event quickstart

sgilda:


sgilda:
@rafabene : Would you like to review? This was the only reference I could find to that link.

There's also a reference to the Bootstrap CSS in the spring-petclinic staticFiles.jsp file, but it's not a link and we do not modify those files: 
    <spring:url value="/webjars/bootstrap/2.3.0/css/bootstrap.min.css" var="bootstrapCss"/>

PR #1734, opened 2016-01-14
Applied workaround for JBEAP-2187 in spring-petclinic

trepel:


sgilda:
@trepel : It looks like a number of JSP source files have changed. I thought we were supposed to work 'out-of-the-box' with no source code changes. Is this a workaround due to a problem with the `petclinic` app or is it due to an issue in JBoss EAP 7?

trepel:
Hi @sgilda , it's difficult to say, there is no comment in https://issues.jboss.org/browse/JBEAP-2187 which would explain why this is happening and I didn't find time to investigate it myself.
The fact is that those relative paths no longer work in EAP7 (at least for me), I think the only thing we can do is to workaround it.

sgilda:
@trepel : We've had a pretty strict rule about not making code changes to the `spring-petclinic` quickstart. We need to find out why this is happening. If it's a bug in JBoss EAP, we should fix it. If the issue is truly with the Spring petclinic code, we need to report it or submit a fix back to the project. If is is their problem and we do need to make a change, we need to document what we changed and why we did it. 

I commented on the JIRA. Let's see what they have to say.

sgilda:
@trepel : I am closing this pull because it looks like they plan to fix the issue.

PR #1736, opened 2016-01-18
JBEAP-2831: can not configure domain in ejb-multi-server QS

wfink:
It seems that the CLI syntax for "thread" and "workmanager" entries has changed.
The CLI commands need to be adjusted

Fix for https://issues.jboss.org/browse/JBEAP-2831


sgilda:
Thanks @wfink ! Merged!

PR #1758, opened 2016-01-26
JBEAP-3041: Updgrade quickstarts to use the resteasy-jackson2-provider

sgilda:


sgilda:
@rafabene : Could you please review these code changes to upgrade the quickstarts to use the new `resteasy-jackson2-provider` included with JBoss EAP 7?

PR #1781, opened 2016-02-12
JBEAP-3386: Remove beans.xml files that are only markers from the qui…

sgilda:
…ckstarts


sgilda:
I am closing this for now. There are too many quickstarts that have issues with this.

PR #1820, opened 2016-03-18
Fill in date first when registering in spring controlleradvice test

trepel:
I came across  "element not clickable" error if using newer version of selenium 2.50.\* in this functional test. The problem is that when filling in date, calendar pops up and hides the "register" button.

Filling in date first and the filling in other info (name, email, ...) is much more stable.


spring-kitchensink-controlleradvice/functional-tests/src/test/java/org/jboss/as/quickstarts/kitchensink/test/page/RegistrationFormPageFragment.java
```diff
@@ -118,6 +112,14 @@ public void register(Member member, boolean fillDate) {
             today.click();
             javascript.executeScript("document.getElementById('datepicker').blur()");
         }
+        nameField.click();
```
sgilda, line 115: @trepel : Did you intend to add the nameField.click()? I assume that is done to put the cursor in the next field, but just want to be sure. :-)

trepel:
Yes, it's intentional. When using miscellaneous versions of Selenium and Firefox, having "nameField.click();" in place proved to be the most stable. It worked everytime.

PR #1829, opened 2016-03-29
[JBEAP-2757] Remove files not relevant to EAP from EAP quickstarts zip

psakar:
no upstream required


sgilda:
Looks good! Merging.

PR #1860, opened 2016-04-20
Removed JBoss Forge version

gastaldi:
JBoss Forge is now at version 3 and is usually known without the version (Forge 1.x is long gone)

Feel free to close this PR if it doesn't make sense


sgilda:
@gastaldi : This makes total sense. I should have removed it long ago. I am merging, but it may be too late to make the JBoss EAP 7 GA release.

Thanks for fixing this!

PR #1898, opened 2016-06-13
JBEAP-4942: Fix and document Arquillian tests for helloworld-mbean qu…

sgilda:
…ickstart


sgilda:
Thanks for the fix @aslak!

PR #1909, opened 2016-06-15
[WIP] Add pipeline definitions for helloworld

tnozicka:
Initial design of our pipeline for helloworld project.

I am aware of few thinks that need to be fixed/changed, but I wanted to share it in early stage.

CC: @ALRubinger, @rupalibehera, @xcoulon

### DIFF
--- a/helloworld/Jenkinsfile
+++ b/helloworld/Jenkinsfile
@@ -1,0 +1,42 @@
+#!/usr/bin/groovy
+node('redhatdistortion-maven') {
+    stage 'Checkout'
+    checkout scm
+
+    stage 'Build'
+    sh 'cd helloworld && mvn clean compile'
+
+    stage 'Run Unit Tests'
+    sh 'cd helloworld && mvn test'
+
+    stage 'Package'
+    sh 'cd helloworld && mvn package'
+
+    stage 'Archive artifact'
+    archive 'helloworld/target/*.war'
+
+    stage 'Create Image'
+    sh 'oc login https://kubernetes.default/ -u openshift-dev -p devel --insecure-skip-tls-verify=true'
+    sh 'oc project helloworld-dev || oc new-project helloworld-dev'
+    sh 'oc process -f https://github.com/tnozicka/openshift-templates/raw/master/binary-build-template.yaml ' +
+       '-v "APP_NAME=helloworld,BUILDER_IMAGE=registry.access.redhat.com/jboss-eap-7/eap70-openshift,OUTPUT_IMAGESTREAM_TAG=helloworld:latest" ' +
+       '| oc apply -f -'
+    sh 'mkdir -p ../artifacts/deployments/ && cp helloworld/target/*.war ../artifacts/deployments/'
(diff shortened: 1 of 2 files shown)


helloworld/Jenkinsfile
```diff
@@ -0,0 +1,42 @@
...
+    sh 'cd helloworld && mvn test'
+
+    stage 'Package'
+    sh 'cd helloworld && mvn package'
+
+    stage 'Archive artifact'
+    archive 'helloworld/target/*.war'
+
+    stage 'Create Image'
+    sh 'oc login https://kubernetes.default/ -u openshift-dev -p devel --insecure-skip-tls-verify=true'
```
jimmidyson, line 19: Hard coded password are not good - can we reuse the service account token in the pod, assuming it has correct permissions?

helloworld/Jenkinsfile
```diff
@@ -0,0 +1,42 @@
...
+    sh 'oc project helloworld-dev || oc new-project helloworld-dev'
+    sh 'oc process -f https://github.com/tnozicka/openshift-templates/raw/master/binary-build-template.yaml ' +
+       '-v "APP_NAME=helloworld,BUILDER_IMAGE=registry.access.redhat.com/jboss-eap-7/eap70-openshift,OUTPUT_IMAGESTREAM_TAG=helloworld:latest" ' +
+       '| oc apply -f -'
+    sh 'mkdir -p ../artifacts/deployments/ && cp helloworld/target/*.war ../artifacts/deployments/'
+    sh 'oc delete builds -l buildconfig=helloworld'
```
jimmidyson, line 25: Deleting all previous builds seems a bit over the top. Shouldn't previous builds be kept?

helloworld/Jenkinsfile
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/groovy
+node('redhatdistortion-maven') {
+    stage 'Checkout'
+    checkout scm
```
jimmidyson, line 4: Shouldn't this be checked out by the build? Does it need to be done again?

helloworld/Jenkinsfile
tnozicka, line 25: It's bug in OpenShift IMO: https://github.com/openshift/origin/issues/9301

helloworld/Jenkinsfile
tnozicka, line 19: True. I made it initially just to work with ADB and beacause I am busy a lot with other important issues.. I have been looking into service accounts after your comment and I think I can build in something like `oc --token=$(cat /run/secrets/kubernetes.io/serviceaccount/token) --certificate-authority=/run/secrets/kubernetes.io/serviceaccount/ca.crt ....` hopefully tomorrow.

helloworld/Jenkinsfile
tnozicka, line 4: I get your point. But without that step, it does not work. This might be because of the fact that you can have Jenkinsfile and source from different branches even repositories.

PR #1918, opened 2016-06-30
JBEAP-5148 Updated commons-lang3 to 3.4 and commented out missing sou…

trepel:
…rces jar

I would wait for @MatousJobanek to review this before merging (to be back from PTO on Monday 11th of July. He might think of better solution.

@sgilda If we for some reason can't wait that long, then we should merge this.


sgilda:
Thanks for looking into this @trepel . This can wait for review by @MatousJobanek .

MatousJobanek:
Hi,
unfortunately, this fix won't work - the whole repository http://mirrors.ibiblio.org/maven2/ is probably down as it doesn't contain anything.
I would use different custom repository - for example the apache's one:  https://repo.maven.apache.org/maven2
I'll send a PR. This one can be closed.
Thanks Tomas for your effort :-)

sgilda:
@MatousJobanek : I can fix this one. I'll submit the pull request and ask for your review.

PR #1920, opened 2016-07-11
Add missing line break to template README

bparry02:


sgilda:
Great catch! Thanks @bparry02 !

PR #1967, opened 2016-08-19
Add EmailAddress with UserType registered via package-info.java.

nakulgawande09:
This works in AS 7.1.1 but fails with EAP 6.1, as the package-info.java
is not found.

Specifying the UserType directly via an @Type at the field works with both
jboss versions.


sgilda:
I am sorry, but all the JBoss EAP 6.x branches of the quickstarts are closed and we are no longer taking any updates to it. 
Please download the JBoss EAP 6.4 version of the quickstarts. I believe it works in that verson.

PR #1982, opened 2016-09-01
added support for secure web sockets with the protocol is https

chrisruffalo:
I had the need to test an environment that was running on HTTPS with this quickstart to make sure that WebSockets was being proxied correctly. The quickstart did not work out-of-the-box with this setup so I added this change for it to work correctly.

### DIFF
--- a/websocket-hello/src/main/webapp/index.html
+++ b/websocket-hello/src/main/webapp/index.html
@@ -23,7 +23,8 @@
             var websocket = null;
 
             function connect() {
-                var wsURI = 'ws://' + window.location.host + '/jboss-websocket-hello/websocket/helloName';
+                var wsProtocol = window.location.protocol == "https:" ? "wss" : "ws";
+                var wsURI = wsProtocol + '://' + window.location.host + '/jboss-websocket-hello/websocket/helloName';
                 websocket = new WebSocket(wsURI);
 
                 websocket.onopen = function() {


sgilda:
@chrisruffalo : Thanks for the fix! How can I test this?

Also, fyi, I created this tracking JIRA: https://issues.jboss.org/browse/JBEAP-5849

chrisruffalo:
If you set up an SSL connector (either on EAP/Wildfly or in HTTPD) and
connect with https that would do it.

Sadly it's not that practical to test quickly.

sgilda:
Thanks @chrisruffalo ! Merged the fix.

sgilda:
@chrisruffalo : Do we need to apply fixes to the websocket-client, websocket-endpoint, and thread-racing quickstarts also? See:
- https://github.com/jboss-developer/jboss-eap-quickstarts/blob/7.1.x-develop/websocket-client/src/main/webapp/index.html#L88
- https://github.com/jboss-developer/jboss-eap-quickstarts/blob/7.1.x-develop/websocket-endpoint/src/main/webapp/resources/js/javascript.js#L113
- https://github.com/jboss-developer/jboss-eap-quickstarts/blob/7.1.x-develop/thread-racing/src/main/webapp/index.html#L46

chrisruffalo:
Yes, Chrome especially does not like it when you open a web socket
connection to an endpoint that doesn't match.

sgilda:
@chrisruffalo : I submitted a pull request for websocket-endpoint and thread-racing. I'm holding off on websocket-client because the relay server is not working and I can't test it. Would you mind reviewing that pull request? 
https://github.com/jboss-developer/jboss-eap-quickstarts/pull/1986

PR #1986, opened 2016-09-02
JBEAP-5489: Update protocol for thread-racing and websocket-endpoint quickstarts

sgilda:
Based on thiis pull request [1] from @chrisruffalo , I discovered other cases where the code should be updated to handle secure connections. This fixes 2 of the 3. I am waiting on websocket-client because it is currently not working.

[1] https://github.com/jboss-developer/jboss-eap-quickstarts/pull/1982

### DIFF
--- a/thread-racing/src/main/webapp/index.html
+++ b/thread-racing/src/main/webapp/index.html
@@ -43,7 +43,8 @@
         } else {
             host = window.location.host;
         }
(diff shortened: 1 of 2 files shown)


sgilda:
@chrisruffalo : Would you mind reviewing this pull request for me to make sure it is correct? Thanks!

PR #2001, opened 2016-09-29
JBEAP-6232: bean-validation - add Engish messages to validation annot…

sgilda:
…ations to prevent test failures in other languages


trepel:
@sgilda Hello Sande, I can confirm your changes work. Based on the fact that this quickstart is not intended to demonstrate the localization recommended practices, it looks good to me. Feel free to merge.

sgilda:
@trepel : Thanks for reviewing the pull request. If I find time, I will try to add localization to this quickstart because it would be a good example, but I don't think it's critical. At least is works now in any language. Thanks again!

PR #2007, opened 2016-10-06
7.0.0.GA release

psakar:
content of officially released quickstarts zip


psakar:
@sgilda 7.0.x branch does not contain officially released quickstarts and diverges from Paul's tag. What do you suggest to fix it ?

sgilda:
@psakar : The 7.0.x branch should not contain the officially released quickstarts. That is the purpose of the 7.0.0.GA tag.

I am not clear why you are trying to update the 7.0.x branch. The "7.0.0-SNAPSHOT" is fine. 
- If you look at the 6.2.x, 6.3.x, and 6.4.x branches, you will see that they still contain the "-SNAPSHOT" versions in the POM files. They remain that way in the event that the quickstarts need to be rebuilt at some point.
- When @pgier did the builds, the "-SNAPSHOT" was replaced with the appropriate "-redhat-x" tag for the quickstart ZIP files.
- The releases that are tagged 6.2.0.GA, 6.3.0.GA, 6.4.0.GA, and 7.0.0.GA are the officially released quickstarts.

You need to work with @pgier. He can explain his process better than I can.

This pull request is not valid and I am closing it.

PR #2036, opened 2016-10-20
JBEAP-6344: Do not use injected JMSContext in thread-racing quickstart

sgilda:
Injected JMSContext are not supported when call is coming from a
WebSocket endpoint as they do not define CDI scope for managed bean (as
reported in https://issues.jboss.org/browse/CDI-370 and
https://java.net/jira/browse/WEBSOCKET_SPEC-196).

Instead, look up the default JMS ConnectionFactory and create an
application-managed JMSContext in a try-with-resources block.


sgilda:
@ctomc : Would you like to review?

ctomc (approved):

PR #2040, opened 2016-10-28
JBEAP-6200: Adapt the servlet-security quickstart to use Elytron

sgilda:


sgilda:
@ctomc : Would you like to review this pull request? I tested it and it works great against DR7.

PR #2042, opened 2016-10-31
JBEAP-6717: jaxws-ejb POM file update to remove exclusion of Sun jaxb-api

sgilda:
JBEAP-6717: jaxws-ejb POM file update to remove exclusion of Sun jaxb-api


sgilda:
@ctomc : I'm not sure if this update totally resolves your concerns with this quickstart's dependency issues.

PR #2059, opened 2017-01-23
Patch for 403 Forbidden error - added a web.xml deployment descriptor

sgilda:
file setting up "greet.xhtml" as the main page of the web app.


sgilda:
@ctomc : While reviewing JBEAP-7733, I noticed this fix was applied to the WildFly quickstarts but not to the JBoss EAP quickstarts. Can you let me know if it should be applied? If not, I will close this pull request.

ctomc:
merged and pushed to downstream repository. closing here.

PR #2083, opened 2018-04-16
Generate and add README.html files to the openshift branch

sgilda:


bstansberry:
ejb-in-war/ejb-in-war-keystore.jks is not in the zip and isn't in the 7.1 branch variant of the QS. I suspect it got in this PR by mistake.

bstansberry:
helloworld-ws and tasks are not in this branch so we shouldn't add their README.html

bstansberry:
I've compared all the other README.html files here to the equivalents in the qs zip and they match except for the "Last updated..." text. I think that mismatch is ok.

batch-processing/README.html
```diff
@@ -0,0 +1,883 @@
...
+</div>
+</li>
+<li>
+<p>Click <strong>Create</strong> to create the project.</p>
+</li>
+<li>
+<p>On the <strong>My Projects</strong> page, choose your new <strong>batch-processing</strong> and click <strong>Browse Catalog</strong>.</p>
+</li>
+<li>
+<p>Choose <strong>JBoss EAP 7.1 (no https)</strong> and then click <strong>Next</strong> to view the <strong>Information</strong> page.</p>
```
bstansberry, line 788: These all say "JBoss EAP 7.1 (no https)" instead of "JBoss EAP CD (no https)".

I think https://github.com/jboss-developer/jboss-eap-quickstarts/pull/2082 wasn't in the branch you generated this from.

bstansberry:
I've corrected the various things I mentioned ^^^ in commits in the branch where I'm reconciling the deltas to the zip: https://github.com/jboss-developer/jboss-eap-quickstarts/compare/openshift...bstansberry:toward-zip

batch-processing/README.html
sgilda: Correct. That PR was never merged.

sgilda:
@bstansberry : I cleaned up this PR to remove the keystore and the 2 html files. It still contains the old OpenShift template name. I also created a replacement PR that uses the updated OpenShift template name. See https://github.com/jboss-developer/jboss-eap-quickstarts/pull/2084.

PR #2086, opened 2018-04-18
Openshift

Athani:
ya


sgilda:
This PR is not valid and the 7.1 branch of the quickstarts is final.

PR #2094, opened 2019-01-04
Updated link to all quickstarts

paulrobinson:


paulrobinson:
Pretty sure this will be out of date now.

PR #2100, opened 2019-09-10
Add Web Archive (WAR) version of ha-singleton example

sqtran:
There are slight differences in how it's packaged for WAR vs JAR, so this is a working example.


emmartins:
outdated

PR #2118, opened 2021-07-29
JBPEAP-22293 new quickstart to show access for RHDG caches

wfink:
https://issues.redhat.com/browse/JBEAP-22293

Adding an example how to access the configured Infinispan cache

### DIFF
--- a/clustered-cache/src/main/java/org/jboss/as/quickstarts/datagrid/eap/subsystem/web/JsfController.java
+++ b/clustered-cache/src/main/java/org/jboss/as/quickstarts/datagrid/eap/subsystem/web/JsfController.java
@@ -1,0 +1,118 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2018, Red Hat, Inc. and/or its affiliates, and individual
+ * contributors by the @authors tag. See the copyright.txt in the
+ * distribution for a full listing of individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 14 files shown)


clustered-cache/src/main/java/org/jboss/as/quickstarts/datagrid/eap/subsystem/web/JsfController.java
```diff
@@ -0,0 +1,118 @@
...
+
+//   @Resource(lookup = "java:jboss/infinispan/container/jdg-container")
+//   CacheContainer container;   // container could be injected if needed
+
+  /**
+   * The direct lookup like
+   *
+   * @Resource(lookup = "java:jboss/infinispan/cache/jdg-container/EAPcache")
+   *
+   * is not working because of https://issues.redhat.com/browse/WFLY-11817.
```
pferraro: I would instead link to the corresponding JBEAP jira:
https://issues.redhat.com/browse/JBEAP-22290

wfink: done

emmartins:
this should instead be contributed to wildfly/quickstart , and have a WFLY JIRA

wfink:
> this should instead be contributed to wildfly/quickstart , and have a WFLY JIRA

more important for the EAP release as it is the first time this is supported.
With WildFly/Infinispan this could be done no matter which release

wfink:
Rebased on top of 7.4.x and tested

emmartins:
@wfink this is not a dev branch or repo, it's just where we publish released EAP quickstart sources, for OpenShift usage mainly. Any development must be done upstream so I suggest that if you still want to propose this then clone/move the JBEAP to WFLY and let's discuss it.

PR #2132, opened 2024-01-09
Update eap-maven-plugin version in pom.xml

tommaso-borgato:


tommaso-borgato:
@jmesnil can you have a look? perhaps I shouldn't have created t in the first place .... anyway it would be useful to upgrade the plugin in order not to bump into https://issues.redhat.com/browse/JBEAP-23890 anymore

jmesnil:
@tommaso-borgato We don't not merge anything in that repo.

Instead we use https://github.com/jbossas/eap-quickstarts and then we push to this repo when we will release EAP 8.

Some of your changes have already been incorporated in that repo: https://github.com/jbossas/eap-quickstarts/blob/c614451c366569e15c16807b3f99c5998686c685/todo-backend/pom.xml#L216-L217 but it seems that we are behind on the build of the eap-maven-plugin: https://github.com/jbossas/eap-quickstarts/blob/c614451c366569e15c16807b3f99c5998686c685/todo-backend/pom.xml#L49C47-L49C47

emmartins:
Hi Tommaso, FYI such update will be done post release, when all Beta sources will be replaced with GA ones. As Jeff said, this is a repo where we don't push fixes directly.

PR #2135, opened 2024-11-07
Update HelloService.java

yutaka-akiyama1:
Edit to SoE ver


emmartins:
hello @yutaka-akiyama1 , please notice that you are sending your changes to ours repository, which I believe is not intended

PR #2144, opened 2025-10-26
Migrate Jboss to Spring boot

tforza:
Outcome of a coding session with Claude Code to review a book.


tforza:
Sorry, this should not have ended up here.

